PR #5, opened 2017-06-12
Refactor config::read_config()

joek13:
Currently, the `config::read_config` method panics if the configuration files are malformed or certain values are missing.

I hope to mitigate this partially by implementing a Config type (which is more readable to crate maintainers and even more descriptive than a tuple) and a ConfigError type for any errors that could occur while loading config. Right now, the only variants of ConfigError are Io and Scan, which correspond to system IO errors (errors reading the config file, etc.) and errors parsing the config file (the config file is not valid YAML). 

One refactor not complete is removing `unwrap`s acting on the parsed YAML. These *will* panic and crash the program if the user enters valid YAML that is missing required information (such as omitting the `servers` section). One possibility is to simply match every config_yaml.get and return an `Err(ConfigError::MissingRequired(String))` if there is a required parameter missing; however, this is needlessly verbose and labor-intensive.

One possibly more elegant solution is to use the `serde_yaml` crate and `serde_derive`. This would allow you to define a Config struct with all the fields (optional or not) and then derive Deserialize and Serialize traits. `serde_yaml` will handle parsing the YAML as well as deserializing Config structs from said YAML. This means that a 100+ line endeavor for loading config files with tons of opportunities to panic could be reduced to potentially less than 50. I think the crate author and maintainers should be the ones to decide which crates to use, though, so I left this particular refactor out.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,12 +3,16 @@
 
 use yaml_rust::Yaml;
 use yaml_rust::YamlLoader;
+use yaml_rust::ScanError;
 
 use std::env::home_dir;
 use std::fs::File;
 use std::io::Read;
 use std::path::{Path, PathBuf};
+use std::error;
+use std::fmt;
 
+#[derive(Debug, Clone)]
 pub struct Config {
     ///List of servers
     pub servers: Vec<Server>,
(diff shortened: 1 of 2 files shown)


joek13:
Whoops, CI failed because I forgot to include `lib.rs` in the last commit.

src/config.rs
```diff
@@ -3,11 +3,35 @@
...
+    ///List of servers
+    pub servers: Vec<Server>,
+    ///Defaults: see definition for Defaults struct
+    pub defaults: Defaults,
+    ///Path to store chatlogs
+    pub logs: PathBuf,
+}
+
+impl Config {
+    pub fn new<P: AsRef<Path>>(servers: Vec<Server>, defaults: Defaults, logs: P) -> Config {
```
osa1: Why not use the owned type as type of `logs` here if you're going to convert it to the owned type anyway? That way you avoid allocations in some cases because if the call site can move it to this function it wouldn't have to pass a reference only to copy it here.

joek13:
> This introduces too much arbitrary formatting changes. If we're going to use rustfmt or something we should do this in a separate commit and do it consistently (i.e. on all files). Could you revert the formatting changes? Let me know if you don't have time or anything.

You're right. My editor was running rustfmt on save without my noticing. I'll revert the formatting changes in another commit.

>Why not use the owned type as type of logs here if you're going to convert it to the owned type anyway? That way you avoid allocations in some cases because if the call site can move it to this function it wouldn't have to pass a reference only to copy it here.

Honestly, I hadn't thought about allocations in that manner. Most of the standard library and other libraries use `AsRef<Path>` for ergonomics. If you're passing it a string, it has to create a new PathBuf anyways, but passing a PathBuf would just copy the PathBuf, resulting in an unnecessary allocation. I wonder if this is common enough to be optimized away. I'll have to do more research, your method may be better.

> I think this may be a good idea. What are the downsides? (other than bloated binaries and longer compile times)

I'm not sure the compiled version of serde_derive vs manually deserializaing actually has a significant difference. Considering the fact that serde_derive is basically implementing the code that deserializes one type in specific, I imagine the derived trait implementation is basically just pulling each field from the source data.

osa1:
Implemented this myself in the meantime: b8d29131e1fad2d3882528eb86c69c024462958d.

PR #12, opened 2017-06-30
Add color themes to the configuration file

umurgdk:
Adds /reload command to load color theme configuration on the fly

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -172,7 +172,7 @@ theme:
 use serde::de::{self, Deserialize, Deserializer, Visitor, MapAccess};
 pub use termbox_simple::*;
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Style {
     /// Termbox fg.
     pub fg: u16,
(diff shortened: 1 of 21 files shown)


src/config.rs
```diff
@@ -72,17 +73,100 @@ defaults:
     auto_cmds: []
 
 # Where to put log files
-log_dir: '{}'", log_dir.as_path().to_str().unwrap())
+log_dir: '{}'
+
+# Color theme based on 256 colors (if supported), colors can be defined as color index (0-255) or with it's name
```
osa1: typo: it's -> its

src/config.rs
```diff
@@ -94,107 +178,188 @@ pub struct Style {
...
-    };
-
-pub const TAB_HIGHLIGHT: Style =
-    Style {
-        fg: 9 | TB_BOLD,
-        bg: 0,
-    };
+const COLORS: [(&'static str, u16); 17] = 
+[
+    // Default bg color of the terminal
```
osa1: This should be default fg, not bg. I'm using a dark color theme and `TB_DEFAULT` renders white.

src/config.rs
```diff
@@ -94,107 +178,188 @@ pub struct Style {
...
-        fg: 5,
-        bg: 0,
-    };
-
-pub const TAB_HIGHLIGHT: Style =
-    Style {
-        fg: 9 | TB_BOLD,
-        bg: 0,
-    };
+const COLORS: [(&'static str, u16); 17] =
```
osa1: Could you add a reference (maybe the wiki link you shared on IRC) here for the mappings?

src/config.rs
```diff
@@ -94,107 +178,188 @@ pub struct Style {
...
+];
+
+const ATTRS: [(&'static str, u16); 3] =
+[
+    ("bold",      TB_BOLD),
+    ("underline", TB_UNDERLINE),
+    ("reverse",   TB_REVERSE)
+];
+
+fn parse_color(val: String) -> Result<u16, ()> {
```
osa1: Why not return `Option<u16>` here?

src/config.rs
```diff
@@ -94,107 +178,188 @@ pub struct Style {
...
+
+        deserializer.deserialize_map(StyleVisitor)
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(default)]
+pub struct Theme {
+    /// Colors used to highlight nicks
+    pub nick_colors: [u8; 14],
```
osa1: I think we shouldn't fix this to 14 colors, users may want to have more or less colors here so maybe we should use a `Vec<u8>`.

One problem with that is we won't be able to use `default_theme()` function for initialization anymore because we can't allocate a vector in a const function.

src/config.rs
```diff
@@ -94,107 +178,188 @@ pub struct Style {
...
+    ("magenta",     13),
+    ("cyan",        14),
+    ("white",       15)
+];
+
+const ATTRS: [(&'static str, u16); 3] =
+[
+    ("bold",      TB_BOLD),
+    ("underline", TB_UNDERLINE),
+    ("reverse",   TB_REVERSE)
```
osa1: I'm thinking maybe we should get rid of "reverse". I can't imagine anyone actually using this..

src/config.rs
umurgdk: I just learned `Option::ok_or_else` and forget to apply XD

osa1:
Hey,

I made the whole TUI parametric on colors used when rendering. We don't have any hard-coded color values and colors should now be changeable in runtime, so /reload command should now work. Nick colors are also held in a Vec now so no need for 14 colors limit. Could you rebase this patch so that I can merge?

umurgdk:
@osa1 updated the branch with latest changes in master. Even though I run tests and successfully run, I couldn't test all the functionality, somehow I can't connect to any IRC server right now (weechat also can't connect, there is something wrong with my workstation)

osa1:
Thanks!

PR #35, opened 2017-11-11
Added /ignore to ignore join/quit messages

### LINKED ISSUE #18: Option to ignore joins/quits
irssi has /ignore which give the option to ignore joins/quits/parts on a per-channel basis. Turning off notifications of joins/quits/parts and all of that altogether would be enough.

Watching all the joins/parts/quits in busy channels can be a bit much, and distracting.

As an aside, I don't want to spam your project with issues for things. If these aren't helpful right now, please let me know.

meain:
Fixes #18

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -410,6 +410,35 @@ impl<'poll> Tiny<'poll> {
             }
         } else if words[0] == "clear" {
             self.tui.clear(&src.to_target());
+        } else if words[0] == "ignore" {
+            match src {
+                MsgSource::Serv { serv_name } => {
+                    self.tui.toggle_ignore(
+                        &MsgTarget::AllServTabs{
+                            serv_name: &serv_name,
+                        }
+                    );
+                }
+                MsgSource::Chan {
+                    serv_name,
+                    chan_name,
+                } => {
+                    self.tui.toggle_ignore(
+                        &MsgTarget::Chan{
+                            serv_name: &serv_name,
+                            chan_name: &chan_name,
+                        }
+                    );
+                }
+                MsgSource::User { serv_name, nick } => {
+                    self.tui.toggle_ignore(
+                        &MsgTarget::User{
+                            serv_name: &serv_name,
+                            nick: &nick
+                        }
+                    );
+                }
+            }
         } else {
             self.tui.add_client_err_msg(
                 &format!("Unsupported command: \"/{}\"", msg),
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -410,6 +410,35 @@ impl<'poll> Tiny<'poll> {
...
+                    chan_name,
+                } => {
+                    self.tui.toggle_ignore(Some(Timestamp::now()),
+                        &MsgTarget::Chan{
+                            serv_name: &serv_name,
+                            chan_name: &chan_name,
+                        }
+                    );
+                }
+                MsgSource::User { serv_name, nick } => {
```
osa1: This part is not quite right .. `MsgSource::User` means that the message has originated from a user tab. So this happens when you run `/ignore` in a private messaging tab. I think in this case we should just apply this to this users's tab, so `MsgTarget::User { serv_name: &serv_name, nick: &nick }` (just replace `AllUserTabs` with `User`).

src/tui/messaging.rs
```diff
@@ -380,33 +384,41 @@ impl MessagingUI {
...
+                    line.add_char('-');
+                    line.set_style(SegStyle::SchemeStyle(SchemeStyle::Faded));
+                    line.add_text(nick);
+                    line.add_char(' ');
+                });
+            }
         }
     }
 
+    pub fn ignore(&mut self, ts: Option<Timestamp>) {
```
osa1: `ts` is unused, please remove it, also from all the call sites.

meain:
@osa1 My bad, fixed it.

osa1:
There's a few things missing:

- What if I disabled notifications in a server, and then run `/join #channel` ? The channel won't inherit the notification state.
- Because of this, `/ignore` in `auto_cmds` won't be useful right now.
- There's no way to see `/ignore` state (true or false) in a tab, so there's no way to tell whether the channel is quiet or `/ignore` is enabled.
- If I have 3 `/ignore`d channels, and 3 non-`/ignore`d, and run `/ignore` in the server tab, I'll still have 3 ignores and 3 non-ignores. This is not really intuitive, I think most UIs would keep `/ignore` state of the server, toggle it, and then set every other tab's `/ignore` state to the server's.

Let's merge this but fix these in follow-up commits, and announce the feature after these (e.g. add to README and CHANGELOG).

Thanks !

meain:
Oh, I didn't think about these because my initial idea was to implement `/ignore` for `MsgSource::Chan` only and then worked on top of that. I will add them in.

PR #39, opened 2017-11-13
Deep create log dir

### LINKED ISSUE #36: tiny crashes when log dir needs deep mkdir
deadstone reported on /r/rust that tiny crashes when the log dir needs deep mkdir. Stack trace:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 2, message: "No such file or directory" } }', /checkout/src/libcore/result.rs:906:4
stack backtrace:
   [insert libstd unwinding here]
  10: <core::result::Result<T, E>>::unwrap
             at /checkout/src/libcore/result.rs:772
  11: tiny::logger::Logger::new
             at src/logger.rs:52
  12: tiny::Tiny::run
             at src/lib.rs:202
  13: tiny::run
             at src/lib.rs:106
  14: tiny::main
             at bin/main.rs:4
  15: __rust_maybe_catch_panic
             at /checkout/src/libpanic_unwind/lib.rs:99
  16: std::rt::lang_start
             at /checkout/src/libstd/panicking.rs:459
             at /checkout/src/libstd/panic.rs:361
             at /checkout/src/libstd/rt.rs:59
  17: main
  18: __libc_start_main
  19: _start
```

BrianOn99:
Fix #36 

I think this is not the best approach.  `fs::create_dir_all` may still fails in many scenario, so returning a `Result` may be better.  But that still need more work.

### DIFF
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -44,7 +44,7 @@ fn init_log_file(file: &mut Write) {
 
 impl Logger {
     pub fn new(log_dir: PathBuf) -> Logger {
-        let _ = fs::create_dir(&log_dir);
+        let _ = fs::create_dir_all(&log_dir);
 
         let debug_logs = {
             let mut log_dir = log_dir.clone();


osa1:
Thanks! Indeed, `logger` module is kind of a mess right now, we should refactor it and remove pretty much all `unwrap()`s because most of them are actually common cases that can happen (e.g. log file removed after tiny started running). I'll try to do that soon.

Note to self: add a line about this in .CHANGELOG.md`.

BrianOn99:
Is there anything I can help?  In particular is there something easy to start with?

osa1:
@BrianOn99 I'm a bit busy for a few days, I think you can help with refactoring the logger. I'll create an issue about this and ping you there.

BrianOn99:
Nice.  I will start looking at it the coming weekend.

PR #40, opened 2017-11-15
Better `/ignore`

meain:
Rest of the fixes from https://github.com/osa1/tiny/pull/35

- [x] What if I disabled notifications in a server, and then run /join #channel ? The channel won't inherit the notification state.
- [x] Because of this, /ignore in auto_cmds won't be useful right now.
- [x] There's no way to see /ignore state (true or false) in a tab, so there's no way to tell whether the channel is quiet or /ignore is enabled.
- [x] If I have 3 /ignored channels, and 3 non-/ignored, and run /ignore in the server tab, I'll still have 3 ignores and 3 non-ignores. This is not really intuitive, I think most UIs would keep /ignore state of the server, toggle it, and then set every other tab's /ignore state to the server's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Connection closure on remote side when TLS is enabled is now handled (#48).
 - `alt-char` bindings implemented to switching between tabs.
 - Fixed some bugs in `join` command used in `auto_cmds` (#49, #38).
+- Tabs can now be moved left/right with `alt-left/right` keys (#52).
+- Input field cursor location now preserved after resize.
+- `TOPIC` messages are now handled (#58).
+- `RPL_AWAY` is now handled (#55). Away message is shown in user tab.
 
 # 2017/11/12: 0.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- `/ignore` command added to ignore `join/quit` messages in channel
+
 # 2017/11/12: 0.3.0
 
 - Fixed a bug that caused wrong scrolling in input field after changing nick.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Unreleased
 
 - `/ignore` command added to ignore `join/quit` messages in channel
+- `/switch` command added to quickly switch to a different tab using a
+  substring of the tab name.
+- `Del` key is now handled. It deletes character under the cursor.
 
 # 2017/11/12: 0.3.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 - `/switch` command added to quickly switch to a different tab using a
   substring of the tab name.
 - `Del` key is now handled. It deletes character under the cursor.
+- Some tweaks and a bug fix (#45) in tab bar rendering. Selected tab is now
+  stays visible in the tab bar after resizing.
+- Connection closure on remote side when TLS is enabled is now handled (#48).
+- `alt-char` bindings implemented to switching between tabs.
+- Fixed some bugs in `join` command used in `auto_cmds` (#49, #38).
 
 # 2017/11/12: 0.3.0
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ serde = "1.0.8"
 serde_derive = "1.0.8"
 serde_yaml = "0.7.1"
 take_mut = "0.2.0"
-term_input = "0.1.2"
+term_input = "0.1.3"
 termbox_simple = "0.2.0"
 time = "0.1"
 
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ colors:
     tab_active:
         fg: default
         bg: default
-        attrs: [underline]
+        attrs: [bold]
 
     tab_normal:
         fg: gray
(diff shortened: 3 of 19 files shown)


meain:
@osa1 How would you like to show the status of `ignore`?

meain:
I have added a simple message on toggle.

osa1:
Thanks, and sorry for late response, I've been a bit busy.

> How would you like to show the status of ignore?

Whatever we do we should keep the noise at minimum and use as little space on the screen as possible.

In addition, tiny should the same when for people who don't use this feature.

What you did (show a simple message on toggle) is fine, but it doesn't show ignore status of a tab. I keep tiny running for long periods of time, sometimes checking once or twice in a day. There's no way for me to understand the ignore status without toggling it twice. Maybe this is fine, I'm not a user of this feature (never used it in other clients), but I feel like we should have an indicator on the screen about ignore status.

Oh, one more problem is that because I can't see ignore status of a server, after `/join #chan`  I don't know ignore status of the channel.

A solution to this is to add a simple ignore status indicator after channel names in the tab bar. By default (ignore is off) everything looks the same. When a channel has ignore on, channel name may look like this: `#rust|i` where maybe `|` and `i` characters have different colors or something like that. Same for server tabs.

This is just one idea. I think it solves the two problems I just mentioned but I'm not sure if it'll look good. Any other ideas @meain ?

meain:
@osa1  How about add something to he left of the nick.
For example, the place where we type would look like:
```
#meain:
```

or 

```
(i)meain:
```

BrianOn99:
> Whatever we do we should keep the noise at minimum and use as little space on the screen as possible.

Agree.  That's what I like most about tiny.

osa1:
- `(i)nick:` looks like `nick` has user mode `i` (especially because `i` is already a user mode that for example freenode automatically sets).

- `#nick:` I don't like this because it adds noise even when you don't use this feature. Also, `#nick` means the channel names `nick` so this is also confusing.

I think we should add some indicator to tab names rather than to nicks to avoid confusion with user modes (see RFC 2812 section 3.1.5).

meain:
@osa1 @BrianOn99 This is how it looks like. Thoughts.
screenshot

meain:
I feel like we are giving a bit more importance to just ignoring joins and quits. If it is like ignoring notifications ( which is some feature I would like to see ) the space used up may be justified.

Maybe its just me.

meain:
> #rust|i where maybe | and i characters have different colors

Making them faded does make a lot more difference than I thought.

meain:
@osa1 ping

BrianOn99:
@meain I don't regularly use `/ignore`, so it is OK if everything not changed when I don't use it.  Let's see what @osa1 think.

osa1:
Sorry for late response,

Two concerns

- `|i` did not turn out too well, I think it looks bad. I'm thinking something else but couldn't come up with anything yet.

- I'm a bit confused about how is this feature useful. Suppose I'm chatting in a channel. When someone that I was talking to leaves there's no way for me to see when `/ignore` enabled. I can disable it, but it's already too late. How is this working on other clients?

meain:
How about this. Highlight currently tab using underline. Non ignored tab using bright colour and  ignored tab using dim color.
@osa1

BrianOn99:
And how about another command `ignore?`, like how vim is use `set filetype?` for checking status?

meain:
Sorry, I got am a bit busy but don't want to leave this incomplete. 

How about the following:
* Ignored tabs will be dimmed, non ignored will be bright, current tab will be underlined. As we don't need to see the char of the current tab to switch I don't think it will clash with `alt+char` bindings.
* implement `/names nick?` to check if they are available

What do you think? @osa1

osa1:
@meain that sounds good.

BrianOn99:
@meain How would `/names nick?` work?

meain:
@BrianOn99 It will show if that nick is available. 
If still available some message like `{nick} is online` else `{nick} left the channel`. Sounds good?

osa1:
> {nick} left the channel

I think `{nick} is not in the channel` is a better message, it more correct in the cases where nick was never in the channel.

meain:
@osa1 Its complete. Just a small issue. Even though `tab_active` was set to use `TB_UNDERLINE` it is being reset by the default config set in `~/.tinyrc.yml` which sets it back to bold. What do you think I should do about that?

meain:
* not ignored

* ignored

meain:
ping @osa1

osa1:
@meain I'll take a detailed look when I have time. One thing I didn't like after a quick look is that tiny looks different now even when I don't use this feature. I'll give more detailed feedback later.

osa1:
@meain Sorry for the delay ... Just one last request and I'll merge this: please make sure that after this patch everything looks the same when `/ignore` is not used. Feel free to add `[i]` suffix to channel names etc. I couldn't come up with a better plan so I'll merge it.

meain:
Tiny now looks the same when not using `/ignore`. When using ignore `[i]` is appended to the right.

meain:
A far fetched plan. How about add a top bar. This screenshot is from `irssi`. It shows the message from channel. We could do something like this and then add the ignore state to right maybe? It could also help with adding more in the future.

osa1:
Thanks!

A top bar may make sense and I agree that it may be a better place to show channel modes (ignore status etc.). Let's maybe sketch out a design for it in a new thread.

PR #42, opened 2017-11-16
Add switch command

meain:
This is something I use personally.

You can now use `/switch` command to switch to the window with a
specific substring.

> Don't know if its he right command name ( `goto` may be better )

Usage:

```
/switch freenode
```

### DIFF
--- a/src/tui/tabbed.rs
+++ b/src/tui/tabbed.rs
@@ -523,6 +523,35 @@ impl Tabbed {
         self.tabs[self.active_idx].set_style(TabStyle::Normal);
     }
 
+    pub fn switch(&mut self, string: &str) {
+        let mut next_idx = self.active_idx;
+        for (tab_idx,tab) in self.tabs.iter().enumerate() {
+            match tab.src {
+                MsgSource::Serv { ref serv_name } => {
+                    if serv_name.contains(string) {
+                        next_idx = tab_idx;
+                        break
+                    }
+                },
+                MsgSource::Chan { ref chan_name, .. } => {
+                    if chan_name.contains(string) {
+                        next_idx = tab_idx;
+                        break
+                    }
+                },
+                MsgSource::User { ref nick, .. } => {
+                    if nick.contains(string) {
+                        next_idx = tab_idx;
+                        break
+                    }
+                }
+            }
+        }
+        if next_idx != self.active_idx {
+            self.select_tab(next_idx);
+        }
+    }
+
     fn next_tab_(&mut self) {
         if self.active_idx == self.tabs.len() - 1 {
             self.active_idx = 0;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- `/switch` command added to quickly switch to a different tab using a substring
+
 # 2017/11/12: 0.3.0
 
 - Fixed a bug that caused wrong scrolling in input field after changing nick.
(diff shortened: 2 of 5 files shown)


src/tui/tabbed.rs
```diff
@@ -523,6 +523,38 @@ impl Tabbed {
         self.tabs[self.active_idx].set_style(TabStyle::Normal);
     }
 
+    pub fn switch(&mut self, string: &str) {
+        for (tab_idx,tab) in self.tabs.iter().enumerate() {
+            match tab.src {
+                MsgSource::Serv { ref serv_name } => {
+                    if serv_name.contains(string) {
+                        self.active_idx = tab_idx;
```
osa1: You need to use `self.select_tab()`, do not manually set `self.active_idx` because it breaks horizontal scrolling in the tab line.

Same for other `self.active_idx = tab_idx;` lines.

src/tui/tabbed.rs
```diff
@@ -523,6 +523,38 @@ impl Tabbed {
...
+                },
+                MsgSource::User { ref serv_name, ref nick } => {
+                    if nick.contains(string) {
+                        self.active_idx = tab_idx;
+                        break
+                    }
+                }
+            }
+        }
+        loop {
```
osa1: You can delete this loop after making the changes in my other comment.

osa1:
Is this really necessary? We already have `alt-N` bindings for switching between tabs.

Every feature means more code we need to maintain and debug, which makes development harder so I'm a bit wary of this kind of patches especially because we already have something like this.

meain:
I just made a PR as it is an option I do enjoy using personally. I tend to have to count every time as tiny does not display numbers ( which I think looks better ). I am fine with closing the PR it if it seems unnecessary.

BrianOn99:
Personally, I would love this feature.

BrianOn99:
When the number of tab grows sometimes it is hard to count.

src/tui/tabbed.rs
```diff
@@ -523,6 +523,38 @@ impl Tabbed {
...
+    pub fn switch(&mut self, string: &str) {
+        for (tab_idx,tab) in self.tabs.iter().enumerate() {
+            match tab.src {
+                MsgSource::Serv { ref serv_name } => {
+                    if serv_name.contains(string) {
+                        self.active_idx = tab_idx;
+                        break
+                    }
+                },
+                MsgSource::Chan { ref serv_name, ref chan_name } => {
```
osa1: `serv_name` unused

src/tui/tabbed.rs
```diff
@@ -523,6 +523,38 @@ impl Tabbed {
...
+                    if chan_name.contains(string) {
+                        self.active_idx = tab_idx;
+                        break
+                    }
+                },
+                MsgSource::User { ref serv_name, ref nick } => {
```
osa1: `serv_name` unused

osa1:
OK, let's do the changes I mentioned in my comments.

@meain could you also add a line to `CHANGELOG.md` and `README.md` (where we explain commands) ?

Thanks.

src/tui/tabbed.rs
```diff
@@ -523,6 +523,35 @@ impl Tabbed {
...
+                },
+                MsgSource::User { ref nick, .. } => {
+                    if nick.contains(string) {
+                        next_idx = tab_idx;
+                        break
+                    }
+                }
+            }
+        }
+        if next_idx != self.active_idx {
```
osa1: :+1:

PR #50, opened 2017-11-29
Remove all prefixes in nicks

### LINKED ISSUE #46: Can't autocomplate op nicks
Channel ops are shown with a `+` character before the nicks and we register those without any modifications. This is causing problems during autocompletion, as `foo<tab>` doesn't complete to `foobar` if `foobar` is an op (i.e. `+foobar`).

BrianOn99:
Fix #46.  This enable autocomplation on those nicks with prefix.

In current state we have no way to know which user is operator.  We may add that feature later, if that is important.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -898,7 +898,7 @@ impl<'poll> Tiny<'poll> {
                             self.tui.new_chan_tab(serv_name, &chan);
                         } else {
                             self.tui.add_nick(
-                                &nick,
+                                drop_nick_prefix(&nick),
                                 Some(Timestamp::now()),
                                 &MsgTarget::Chan {
                                     serv_name: serv_name,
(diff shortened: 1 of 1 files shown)


osa1:
Thanks! I think (but I'm not sure) this should be done in one more place, when we get a JOIN message from the server (search for `tui.add_nick` and you'll see one more call site). Could you check the RFC or modern.ircdocs.horse and see if we need to update that code too?

BrianOn99:
I think you are correct, but it is not memtioned in rfc.  I will join multiple channels these days and hope some op will JOIN.  Otherwise I will dig the code of irssi.  Be prepared to see me in #rustc #ruby #haskell #redox

osa1:
@BrianOn99  I think it doesn't hurt to do the same when we get a `JOIN` message too. I'm going to merge this if you update the code.

BrianOn99:
Ok

BrianOn99:
QUIT and PART may also have this issue, so I think it is better to check them too.  Any idea about them?

osa1:
> QUIT and PART may also have this issue, so I think it is better to check them too. Any idea about them?

Hmm this makes sense. I think we can implement a function like `utils::drop_nick_prefix` and call it before adding/removing nicks.

BrianOn99:
I installed unrealircd server locally for testing.  It turns out PART and QUIT does not have an @ prepended to an operator's nick, even though NAME have it prepended.  No time to test JOIN yet.

BrianOn99:
@osa1 I have not find a way to JOIN as an operator (can only do it with message MODE), but as it does not hurt, now JOIN drop the prefix anyway.

As mentioned before, QUIT and PART have no problem.  And I digged the xchat source code, to my understanding something like "drop prefix" is only called when nick is added (both JOIN and NAME), so I don't handle them and not moved the code to utils.

As an aside I found kicked away from a channel seems not to be handled be tiny.  Is it?

osa1:
Thanks!

> As an aside I found kicked away from a channel seems not to be handled be tiny. Is it?

Correct, `KICK` messages are not handled. Could you open an issue about this?

BrianOn99:
Thank you!

PR #70, opened 2017-12-27
Aur pkgbuild

wrycode:
@osa1 this is my first time using git. Hopefully I am doing this correctly.

I will to change the PKGBUILD's `# Maintainer: Nick Econopouly <wry@mm.st>` line to your contact information after making this pull request, just to see how Github handles the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,11 @@ tiny is an IRC client written in Rust.
 Install the Rust nightly toolchain, clone the repo, and run `cargo install` (or
 `cargo install --force` if you already have an older version installed).
 
+Arch Linux users can use this
+[PKGBUILD](https://github.com/osa1/tiny/tree/master/pkg/archlinux/PKGBUILD)
+to install tiny as a package. Download it into an empty directory, run
+`makepkg` and install the resulting .pkg.tar.xz with `sudo pacman -U`.
+
 Since version 0.3.0 tiny needs OpenSSL or LibreSSL headers and runtime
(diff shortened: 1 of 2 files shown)


BrianOn99:
Just in case you will submit this aur, would `cargo +nightly build` be better?  Otherwise user's default toolchain might be changed.

BrianOn99:
I can verify that this PKGBUILD works, but on Arch we needs `pkg-config` installed first.  It is missing in `makedepends`.

wrycode:
@BrianOn99 thank you, I will update the cargo command and remove the rustup commands.

> I can verify that this PKGBUILD works, but on Arch we needs pkg-config installed first. It is missing in makedepends

`pkg-config` is a member of the `base-devel` group, and therefore should not be included according to the Arch Wiki's
[PKGBUILD](https://wiki.archlinux.org/index.php/PKGBUILD#makedepends) page.

BrianOn99:
Nice catch.

osa1:
Thanks!

Just a comment, I think maintainer should point at you or @BrianOn99 because I have no idea how to maintain this, I don't even use Arch Linux :) Could you update that?

If none of you want to maintain this then it's fine, just let me know.

BrianOn99:
@wrycode Feel free to become the maintainer!

wrycode:
I'm happy to maintain it.

osa1:
Thanks!

Issue #65, opened 2017-12-23
Consider distributing static binaries


osa1:
Per @reading123's request in #61,

I think producing static binaries is a good idea. tiny is light on dependencies (serde being the largest one) and the only culprit for Linux seems to be OpenSSL (or LibreSSL) which is required by `native_tls`. I actually managed to compile tiny using `x86_64-unknown-linux-musl` target, but (1) the binary didn't work (some kind of loader problem) (2) it still had `libssl.so.1.1` and `libcrypto.so.1.1` as dynamic dependencies.

Need to investigate further. It might help to find another Rust project that (1) generates static binaries (2) uses `native_tls` crate.

BrianOn99:
If it is found very difficult to overcome these problems, we may also consider [appimage](https://appimage.org/) or [flatpak](https://flatpak.org/).  From my experience both works decently on Arch and ubuntu.

saethlin:
To get openssl to link statically, just don't let the build script see the dynamic libraries. I downloaded openssl-1.1.0g from [here](https://www.openssl.org/source/), and built them myself. Then I made a directory inside the source one called `lib` and moved the static libraries into that. Then you can build tiny with `OPENSSL_DIR=/home/username/openssl-1.1.0g/ cargo build` and you'll have no openssl dependency at link time.

The much harder problem I've run into is getting the allocator to link properly. I get a huge pile of undefined symbols, all ending in `_chk`. As far as I can tell from reading [this and related issues](https://github.com/briansmith/ring/issues/409) the issue is that these `__memcpy_chk` are source fortification that I think is inserted by gcc to protect against mistakes with the more dangerous bits of the C API. It looks like `ring` decided to discard the fortification for musl builds. Maybe we should do the same? Maybe with a fork of openssl-sys?

saethlin:
I have a solution.

Download the openssl-1.1.0g sources. Configure with `CC=clang ./config no-async`, then make. Move the static lib files to a `lib` directory inside the source directory. Then build tiny with `OPENSSL_DIR=/openssl/source/path cargo build --target=x86_64-unknown-linux-musl`.

`ldd` tells me that the output binary is not a dynamic executable, and it's 7.1 MB and compresses down to 2.6 MB with debug symbols.

Can anyone confirm that this works on other machines and that the binary is actually portable?

osa1:
It worked! Thanks @saethlin .

[...]

PR #74, opened 2018-01-05
Added script to build portable binaries

wrycode:
#65

### DIFF
--- a/pkg/bin/build.sh
+++ b/pkg/bin/build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# TODO:
+# cross-platform dependancy check: clang, perl, ca-certificates (?) others?
+openssl_release="OpenSSL_1_1_0g"
+
+
+wget https://github.com/openssl/openssl/archive/$openssl_release.tar.gz &&
+tar -xf $openssl_release.tar.gz &&
+cd openssl-$openssl_release &&
+CC=clang ./config no-async &&
+make &&
+mkdir lib &&
+cp libcrypto.a libssl.a lib/ &&
+cd ../../.. &&
+OPENSSL_DIR=$PWD/pkg/bin/openssl-$openssl_release/ cargo +nightly build --release
+
--- a/pkg/bin/build.sh
+++ b/pkg/bin/build.sh
@@ -11,6 +11,6 @@ CC=clang ./config no-async &&
 make &&
 mkdir lib &&
 cp libcrypto.a libssl.a lib/ &&
-cd ../../.. &&
-OPENSSL_DIR=$PWD/pkg/bin/openssl-$openssl_release/ cargo +nightly build --release
+cd ../.. &&
+OPENSSL_DIR=$PWD/pkg/bin/openssl-$openssl_release/ cargo +nightly build --release --target=x86_64-unknown-linux-musl
 
--- a/src/tui/messaging.rs
+++ b/src/tui/messaging.rs
@@ -424,12 +424,9 @@ impl MessagingUI {
         }
     }
 
-    /// Set or toggle ignore state.
-    ///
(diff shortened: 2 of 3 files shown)


pkg/bin/build.sh
```diff
@@ -0,0 +1,16 @@
...
+
+wget https://github.com/openssl/openssl/archive/$openssl_release.tar.gz &&
+tar -xf $openssl_release.tar.gz &&
+cd openssl-$openssl_release &&
+CC=clang ./config no-async &&
+make &&
+mkdir lib &&
+cp libcrypto.a libssl.a lib/ &&
+cd ../../.. &&
+OPENSSL_DIR=$PWD/pkg/bin/openssl-$openssl_release/ cargo +nightly build --release
```
osa1: We need `--target=x86_64-unknown-linux-musl` here, otherwise it won't be a statically linked binary.

pkg/bin/build.sh
```diff
@@ -0,0 +1,16 @@
...
+
+
+wget https://github.com/openssl/openssl/archive/$openssl_release.tar.gz &&
+tar -xf $openssl_release.tar.gz &&
+cd openssl-$openssl_release &&
+CC=clang ./config no-async &&
+make &&
+mkdir lib &&
+cp libcrypto.a libssl.a lib/ &&
+cd ../../.. &&
```
osa1: I think this should be `cd ..`, you have one `cd` into `openssl-$openssl_release` so you need one `..`.

pkg/bin/build.sh
osa1: Also, `OPENSSL_DIR` should be `$PWD/openssl-$openssl_release`.

osa1:
This works but depends on where I'm running it from (e.g. running it from top-level doesn't work, I have to `cd` into `pkg/bin`).

Also, we should pass a `--target=...` parameter to `cargo build`.

I wonder if there's an easy-to-use build system for this kind of stuff. It'd be great if we could cache OpenSSL builds, check for clang etc. availability easily etc.

wrycode:
I accidentally stashed a commit, before fetching and rebasing, so I had to do a merge to get my stashed commit to push. Github is now not showing one of your comments, but I fixed the cd command (this is such a learning process...).

Earlier, I couldn't get it working with `--target=x86_64-unknown-linux-musl` and then it started working inexplicably. I should have double checked, it seems that I somehow removed that part of the line with a git operation or something. If I try to run it with `--target=x86_64-unknown-linux-musl`, I get the following errors:
```
   Compiling lazycell v0.5.1
   Compiling foreign-types-shared v0.1.1
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed

error: aborting due to previous error

error: Could not compile `foreign-types-shared`.
warning: build failed, waiting for other jobs to finish...
error[E0463]: can't find crate for `std`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed

error: aborting due to previous error

error: Could not compile `lazycell`.

To learn more, run the command again with --verbose.

```

I don't know how to cross-compile with Rust from Arch Linux. I will look into it, but the command does work for you and [@seathlin ](https://github.com/saethlin), right?


> This works but depends on where I'm running it from (e.g. running it from top-level doesn't work, I have to `cd` into `pkg/bin`).
>
> Also, we should pass a `--target=...` parameter to `cargo build`.
>
> I wonder if there's an easy-to-use build system for this kind of stuff. It'd be great if we could cache OpenSSL builds, check for clang etc. availability easily etc.

 
Does this need to work for more than one developer(you, when you make the binary releases)?

I'm not familiar with build systems. The features you mentioned shouldn't be hard to write in bash but there may be a better way to do this.

PR #80, opened 2018-01-08
Added dbus dependancy to Arch Linux PKGBUILD

wrycode:


osa1:
Thanks! We should merge this after #78.

osa1:
@meain already added this in #78. Closing.

PR #87, opened 2018-02-01
Contributors

wrycode:
I only saw five contributors, let me know if there's a sixth and I'll add them!


osa1:
Thanks! Could you add @wammies too? They committed https://github.com/osa1/tiny/commit/23e8905a2d1a8a94c48bf9d0fc165caf454b483e but because of a git configuration error on their side the name isn't in Github's contributor list.

osa1:
Merged!

PR #98, opened 2018-04-11
Aur pkgbuild

### LINKED ISSUE #73: Add tiny to the AUR
Just about to try this - would be nice if it were in the AUR instead of just a misc package.

wrycode:
This resolves #73. When binaries are available the AUR package will be called tiny-irc-client, as there's already a [tiny](https://aur.archlinux.org/packages/tiny).


osa1:
Awesome! Thanks!

PR #102, opened 2018-08-11
Notification type show in the tab bar

meain:
How do you wish to have the notification mode shown in the tab bar?

I have a simple implementation here. Some more changes to make but does this look good?
Ideally I think we should go with a top bar, but I don't think I can pull that off by myself.

image

### DIFF
--- a/src/tui/tab.rs
+++ b/src/tui/tab.rs
@@ -57,7 +57,7 @@ impl Tab {
     pub fn width(&self) -> i32 {
         // TODO: assuming ASCII string here. We should probably switch to a AsciiStr type.
         self.visible_name().len() as i32 +
-            if self.widget.get_ignore_state() { 0 } else { 3 }
+            if self.widget.get_ignore_state() { 0 } else { 3 } + 3
     }
 
     pub fn draw(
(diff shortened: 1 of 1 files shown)


osa1:
Thanks for doing this! I wonder how to do this with minimal amount of extra noise on the screen.

To me notify settings are not something you change too much, and the current settings is not important enough that you want to see it on the screen always. Also, using 3 character per tab for this seems like too much and makes the tab bar significantly larger.

I wonder how other IRC clients do this. Any ideas?

Out of curiosity, do you really need to see notification setting for each channel on the screen always?

meain:
I don't personally need this feature, but saw an issue for this and though someone might need it.

Yeah, it makes the tab a bit too long.
If it where upto me I would only have it if we had atop bar and move it to the top.

osa1:
> I don't personally need this feature, but saw an issue for this and though someone might need it.

Well, if no one actually needs then no need to add more code which costs us time and energy as we have to write, debug and maintain that code!

I think a top-bar makes more sense here, and we could make it toggle-able to avoid noise. For example, the bar would be only visible when alt is pressed etc. It could also show channel title, and maybe some more information like number of users in the channel.

Implementation-wise I think as simple overlay over `msg_area` in `MessageingUI` would work. We'd first draw `msg_area`, then draw the overlay, so that the overlay is visible over the `msg_area`.

Would you be willing to give it a go? Let me know if you need any help.

meain:
Sorry, but I don't think I will have enough time to look into it. If someone else is willing to take it up, feel free.

meain:
Basic header implementation in PR #109. Closing this PR.

PR #104, opened 2018-08-15
removed features which are not required anymore

nifker:


nifker:
For compilation on my machine I changed the origin of the crate 'term_input'.

osa1:
Thanks!

We need to find a way to allow changes that touch multiple packages in PRs. I'll open an issue about this.

PR #107, opened 2018-08-19
Use relative paths of term_input and termbox_simple in Cargo.toml

osa1:
This should fix CI. As far as I understand I should still be able to do
`cargo publish` as long as those packages are also published (or maybe
`cargo publish` publishes those for me if they're not? I'm not sure).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,8 +33,8 @@ serde_derive = "1.0.8"
 serde_yaml = "0.7.1"
 take_mut = "0.2.0"
 tempfile = "3.0.3"
-term_input = { path = "term_input" }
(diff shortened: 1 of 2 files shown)


osa1:
It turns out we can specify both a relative path and a version number:

```
term_input = { path = "term_input/", version = "..." }
```

Perhaps this is what we're supposed to do here. Easy to test but I don't want to publish a package just to test this..

osa1:
Indeed that's what we need to do, as explained in https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories . Update coming.

PR #126, opened 2019-02-22
updated Arch Linux package and removed dated instructions 

wrycode:
from the README, the PKGBUILD now works out-of-the-box for all users but contains instructions for users with a preferred method of installing the Rust nightly toolchain. I updated the version on the AUR too.


osa1:
Thanks!

Issue #136, opened 2019-09-19
async_refactor branch checklist


osa1:
Stuff to be done before release:

- [x] Finish logging
- [x] Address name resolving step should be interruptible -- currently I can't do e.g. `/connect` if the task is blocked on `getsockaddr()`. Similarly I can't quit.
- [x] SASL authentication
- [x] Implement sending pings on timeout.
- [x] Reconnecting after a disconnect does not work.
- [x] `C-c` does not really quit -- it closes the connections (I think, at least it sends QUIT messages) but the TUI hangs.
- [x] Port 1d2a690
- [x] ~~Find a way to disable stderr logging in release build~~ Use env_logger for debug logs

osa1:
Almost done!

osa1:
There are still a few `await`s in the client main loop that blocks the client; those should be refactored:

- [ ] The `addr_iter` await call
- [ ] The delay when address name resolve fails
- [ ] The delay when connection fails (after address name resolving)

Also, it seems like we currently try to reconnect immediately when connection fails *after* it's established. This is different than the previous behavior where we waited for 30 seconds or so.

osa1:
We have PR now and we'll track progress there.

PR #138, opened 2019-09-29
Rewrite for better code structure, lots of refactoring, use tokio/async

### LINKED ISSUE #3: Address name resolver blocks the UI
The code for initializing `TcpStream`s looks like this:

```rust
fn init_stream(serv_addr: &str, serv_port: u16) -> TcpStream {
    let stream = TcpBuilder::new_v4().unwrap().to_tcp_stream().unwrap();
    stream.set_nonblocking(true).unwrap();
    // This will fail with EINPROGRESS
    let _ = stream.connect((serv_addr, serv_port));
    stream
}
```

Here `connect()` first calls `to_socket_addrs()` which does address name resolving by calling `getaddrinfo()`, which in turn blocks the thread.

One alternative is to use `getaddrinfo_a()`, but I'm not sure how portable this would be.

Another alternative is to use [futures](https://github.com/alexcrichton/futures-rs) to run `getaddrinfo()` and poll the results in tick handlers.

### LINKED ISSUE #56: Private messages not logged
Only raw bytes and channel messages are logged

### LINKED ISSUE #94: Feature: use an IRC library, or refactor IRC protocol logic out of tiny for consumption by others

### LINKED ISSUE #100: Finally fix the logging, release 1.0
I think we should finally fix the logging and release 1.0.

The problem is we currently don't log all the info we want (e.g. privmsgs, see #56). I think a logger function with type `fn(&Conn, &ConnEv)` would be really helpful to implement all of the logging logic in one place. Should not be too hard.

In the future I want to be able to show logs in the UI (e.g. when a privmsg tab is created with a user previous messages would be shown in the tab), but that's a lot of work and needs some more detailed design. Current design has the advantage that the log files are readable plain-text files.

### LINKED ISSUE #132: Crash when log_dir: points to an inaccessible location
When I try to run `tiny` on macOS 10.15 (19A512f), it panics with this output:

```
Any mentions to you will be listed here.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', src/libcore/result.rs:1084:5
mentions                                                                                                                                                                    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Running with backtrace gives me:

```
Any mentions to you will be listed here.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', src/libcore/result.rs:1084:5
mentions                                                                                                                                                                    stack backtrace:
                                                                                                                                                                                               0: std::panicking::default_hook::{{closure}}
                           1: std::panicking::rust_panic_with_hook
                                                                     2: std::panicking::continue_panic_fmt
                                                                                                             3: rust_begin_unwind
                                                                                                                                    4: core::panicking::panic_fmt
                                                                                                                                                                    5: core::result::unwrap_failed
                                                                                                                                                                                                     6: tiny::Tiny::run
       7: tiny::main
                       8: std::rt::lang_start::{{closure}}
                                                             9: main
```

I built `tiny` with rustc 1.38.0-nightly (d3f8a0b5d 2019-08-04) from commit 35752a3.

osa1:
This PR basically rewrites the whole app except the TUI parts. The main motivation is to make the code easier to work with.

Main changes are:

- The code is now split into a few crates, with few inter-dependencies. Dependencies between the new crates look like this:

      tiny
      ├── libtiny_client
      │   ├── libtiny_logger
      │   │   └── libtiny_ui
      │   └── libtiny_wire
      ├── libtiny_logger
      ├── libtiny_tui
      │   └── libtiny_ui
      ├── libtiny_ui
      └── libtiny_wire

- The main crate (tiny, which previously implemented pretty much the whole program) now does a few things:

  - Updates the UI on events from a client (libtiny_client). This is implemented in `conn.rs` in 510 lines.
  - Updates IRC clients on events from the UI (currently libtiny_tui). This is implemented in `ui.rs` in 159 lines.
  - Implements commands like `/connect`, `/away`, `/join` etc. (UI-related commands are now handled in the UI libraries, e.g. libtiny_tui)

  Other than these it just glues the libraries together: initializes TUI, clients, logger etc.

- libtiny_wire implements parsing and generating IRC messages. Only difference from the previous `wire` module is `QUIT` and `NICK` messages now have a field for channels of the user who quit or changed their nick. libtiny_client updates these fields, to be used by the main crate.

- libtiny_client implements the IRC driver. It handles reconnections, nick selection, authentication, ping/pongs, sending messages etc. Users interact with a client using a the `Client` methods and receives IRC events from a tokio `mpsc` channel.

  Note that once created, a client is only destroyed when the user wants to close the connection to the server -- that is, clients are not destroyed on connection errors.

  A client now maintains a list of nicks for every channel joined, and updates `chans` fields of `QUIT` and `NICK` messages before sending them to the user, so that UIs or the logger doesn't have to maintain a list of channels a user is in.

- libtiny_ui defines a trait for UIs and a few utilities for combining two UIs (see below).

  (This will be used for a GUI in the future, see the placeholder libtiny_gui crate)

- libtiny_tui is the good old TUI implementation, which now implements the UI trait from libtiny_ui. The top-level is refactored to support use with tokio -- most notably it implements `Clone` and methods now require `&self` (instead of `&mut self` as before).

  Similar to `Client` (libtiny_client), TUI events are returned to the user using a mpsc channel.

  Other than that the code is the same as before.

- libtiny_logger implements logging. Logger is implemented as a UI, so in the use site we don't actually do any logging, we just combine the main UI (the TUI) with a logger (if logging is enabled by the user), and then rest of the code just updates the UI as usual. This is quite convenient in the use sites.

For debug logging we now use `env_logger`. By default debug logging is disabled. When enabled logging is done to stderr. Because some of the dependencies (e.g. tokio) also use the standard `log` crate it's a good idea to limit the logging scope, using something like:

    RUST_BACKTRACE=1 RUST_LOG="libtiny=debug,tiny=debug" tiny 2>stderr

Breaking changes
----------------

The only breaking change is that the `/reload` command is currently disabled. This is because the command is somewhat special in that it requires parsing the config file in the UI, and I couldn't find a good way to implement this yet.

Other than that this is pretty much just refactoring. Only user-facing change is the new logger (it's a good idea to remove the old logs directory before starting using new tiny, but it's not necessary, logging will still work).

Trivia
------

- Code size: before 7498 loc, after 7762 loc (+264 lines, +3.5%)
- Debug executable size: before 24M, after 54M (+30M, +125%)
- Release executable size: before 2.3M after 3.9M (+1.6M, +69%)
- Debug build time after `git clean -xfd`: before 17.90s, after 26.96s (+9.06s, +50%)
- Release build time after `git clean -xfd`: before 39s, after 1m11s (+32s, +82%)
- Number of transitive dependencies (using `cargo metadata --format-version=1`): before 118, after 168 (+50, +42%)

---

Fixes #100
Fixes #94
Fixes #3
Fixes #56 
Fixes #132


osa1:
There are a few remaining issues in libtiny_client before merging this, see #136 .

lzutao:
This helps https://github.com/rust-lang/rust/pull/65819

PR #143, opened 2019-10-03
Implement Termbox cell API using termion

### LINKED ISSUE #115: Some unicode characters cause glitches
I'll post a screenshot next time this happens, but basically some unicode characters cause glithces that even a `tb_clear()` doesn't fix.

What happens that the unicode character stays visible in its original location when you scroll (either with `shift+arrow` or as new messages appear), and even changing tabs don't make the character go away.

Only way to fix the glitch that I could find is to force a resize.

Probably termbox related.

osa1:
This ports termbox's "cell" abstraction to Rust, removing the last piece
of C code from the repo.

Input and event handling of the original termbox was removed long time
ago, and we've been removing bits and pieces from the original termbox
since then, leaving only the bits we need. The remaining pieces was
basically just the cell abstraction, which we implement in Rust with
this patch, getting some help from termion.

Performance is on par with the original C implementation: libtiny_tui's "bench"
example, when passed a 19M file, runs in 20.1s in C implementation, and 20.2s in
Rust implementation. C implementation writes 918,141,709 bytes to stdout, Rust
implementation write 915,387,921 bytes (-0.2%). The difference may either be
better cursor location caching in the Rust version or a bug. So far I'm unable
to find any rendering bugs, so I'm inclined to think the former.

Co-authored-by: Emilio González <emigr2k1@gmail.com>


osa1:
> I noticed some slowness when using this branch: mainly when I run LIST in Freenode I think the command takes much longer to run, probably not because of networking but because the TUI is slower now. I should implement some rendering benchmarks and compare this branch with master.

Actually it's not slow at all in release mode. I guess debug and release mode performance changes more now because the library was previously implemented in C (and debug/release mode made less difference in generated code) but not it's in Rust (where release mode makes huge difference).

osa1:
Attached file is output of `cargo flamegraph` after running `LIST` in freenode. TLDR is that tiny spends 65.37% of the time in `Termbox::present()`, and most of it is in `std::io::Write::write_fmt` (51.49% of all runtime). I think this should be improved by replacing `write!` calls with `extend_from_slice`s.

[flamegraph.tar.gz](https://github.com/osa1/tiny/files/3689360/flamegraph.tar.gz)

osa1:
Unfortunately simply using `extend_from_slice` instead of `write!` does not work great as we have to generate `String`s for `termion` structs anyway (e.g. `Goto`) -- `termion` doesn't provide a way to directly encode a struct into a given buffer.

Diff:

```diff
diff --git a/termbox/src/lib.rs b/termbox/src/lib.rs
index f9e1d48..95a7d4c 100644
--- a/termbox/src/lib.rs
+++ b/termbox/src/lib.rs
@@ -242,18 +242,14 @@ impl Termbox {
                 if cw > 1 && (x + (cw - 1)) >= usize::from(self.front_buffer.w) {
                     // Not enough room for wide ch, send spaces
                     for i in x..usize::from(self.front_buffer.w) {
-                        write!(
-                            self.output_buffer,
-                            "{} ",
-                            termion::cursor::Goto(i as u16 + 1, y as u16 + 1)
-                        )
-                        .unwrap();
+                        self.output_buffer.extend_from_slice(termion::cursor::Goto(i as u16 + 1, y as u16 + 1).to_string().as_bytes());
+                        self.output_buffer.push(b' ');
                     }
                 } else {
+                    self.output_buffer.extend_from_slice(termion::cursor::Goto(x as u16 + 1, y as u16 + 1).to_string().as_bytes());
                     write!(
                         self.output_buffer,
-                        "{}{}",
-                        termion::cursor::Goto(x as u16 + 1, y as u16 + 1),
+                        "{}",
                         back_cell.ch
                     )
                     .unwrap();
@@ -273,12 +269,7 @@ impl Termbox {
         }
 
         if let Some((x, y)) = self.cursor {
-            write!(
-                self.output_buffer,
-                "{}",
-                termion::cursor::Goto(x + 1, y + 1),
-            )
-            .unwrap();
+            self.output_buffer.extend_from_slice(termion::cursor::Goto(x + 1, y + 1).to_string().as_bytes());
         }
 
         self.tty.write_all(&self.output_buffer).unwrap();
@@ -306,23 +297,13 @@ impl Termbox {
             Some((x, y)) => match self.cursor {
                 None => {
                     self.cursor = Some((x, y));
-                    write!(
-                        self.output_buffer,
-                        "{}{}",
-                        termion::cursor::Goto(x + 1, y + 1),
-                        termion::cursor::Show
-                    )
-                    .unwrap();
+                    self.output_buffer.extend_from_slice(termion::cursor::Goto(x + 1, y + 1).to_string().as_bytes());
+                    self.output_buffer.extend_from_slice(termion::cursor::Show.as_ref());
                 }
                 Some((x_, y_)) => {
                     if x != x_ || y != y_ {
                         self.cursor = Some((x, y));
-                        write!(
-                            self.output_buffer,
-                            "{}",
-                            termion::cursor::Goto(x + 1, y + 1)
-                        )
-                        .unwrap();
+                        self.output_buffer.extend_from_slice(termion::cursor::Goto(x + 1, y + 1).to_string().as_bytes());
                     }
                 }
             },
@@ -388,21 +369,11 @@ impl Termbox {
         }
 
         if fg != 0 {
-            write!(
-                self.output_buffer,
-                "{}",
-                termion::color::Fg(termion::color::AnsiValue(fg)),
-            )
-            .unwrap();
+            self.output_buffer.extend_from_slice(termion::color::Fg(termion::color::AnsiValue(fg)).to_string().as_bytes());
         }
 
         if bg != 0 {
-            write!(
-                self.output_buffer,
-                "{}",
-                termion::color::Bg(termion::color::AnsiValue(bg))
-            )
-            .unwrap();
+            self.output_buffer.extend_from_slice(termion::color::Bg(termion::color::AnsiValue(bg)).to_string().as_bytes());
         }
     }
 }
```

New flamegraph is attached: 63.11% in `Termbox::present`, 37.35% in `to_string()`.
[flamegraph.tar.gz](https://github.com/osa1/tiny/files/3689464/flamegraph.tar.gz)

osa1:
Flamegraph for the master branch: 
[flamegraph.tar.gz](https://github.com/osa1/tiny/files/3689523/flamegraph.tar.gz)

TLDR: It's so much faster.

osa1:
Using the benchmark added in 2b18d91 I compared this branch with master: for the file that master takes 8 seconds to show this branch takes 45 seconds ... so yeah, this is slow.

osa1:
Too bad this is too slow; it'd be good to remove C code and be 100% Rust.

Maybe someone more knowledgeable in Rust than me can figure this out.

emigr2k1:
@osa1 can you try [my branch](https://github.com/emigr2k1/tiny/tree/termbox_termion)? In release my bottleneck is now the syscalls to print. Unless I'm profiling wrong.

I only removed the write!s and make it print the cursor directly to the output buffer.

osa1:
Thanks @emigr2k1 -- just tried your solution. In master branch the `bench` example (in libtiny_tui) when passed a large file runs in 4 seconds, your branch runs in 35 seconds.. so it's still quite slow.

emigr2k1:
@osa1, yeah, I just wanted to point out that before the bottleneck was the write! but at least for me now, the bottleneck is the syscall.

The whole "present" function in C termbox runs in less than 1 millisecond. In the Rust termbox, just doing self.tty.write_all(...) runs in a few milliseconds. I also tried:
`        
unsafe {
            use std::os::unix::io::AsRawFd;
            use libc::c_void;
            libc::write(self.tty.as_raw_fd(), self.output_buffer.as_ptr() as *const c_void, self.output_buffer.len());
        }
`
but it is quite slow too. I'm getting 47% of the time spent on the kernel.

osa1:
@emigr2k1 Interesting.. I wonder if we're writing more stuff to the output buffer in the Rust version. Both versions use a buffer to reduce syscalls, and they should be doing the same amount of syscalls, writing same data, but maybe I did something slightly different in my port and we're currently writing more stuff to the buffer. I should check this.

emigr2k1:
I'll check it too. I just benchmarked it, and I'm getting around 3-5 ms for the write syscall, and 0 ms for the rest of the "present" function.

osa1:
OK so if I'm counting this right for the same benchmark master branch flushes (in total, not in one "present" call) 192,489,158 bytes, and this branch flushes 2,011,204,522. That's more than 10x.

The changes I made on master branch

```
diff --git a/termbox/cbits/termbox.c b/termbox/cbits/termbox.c
index aeaedeb..cf4763d 100644
--- a/termbox/cbits/termbox.c
+++ b/termbox/cbits/termbox.c
@@ -12,6 +12,7 @@
 #include <termios.h>
 #include <unistd.h>
 #include <wchar.h>
+#include <inttypes.h>
 
 #include "termbox.h"
 
@@ -64,6 +65,8 @@ static void send_clear(void);
 /* may happen in a different thread */
 static volatile int buffer_size_change_request;
 
+static uint64_t total_flushed = 0;
+
 /* -------------------------------------------------------- */
 
 int tb_init(void)
@@ -112,6 +115,8 @@ int tb_init(void)
 
 void tb_shutdown(void)
 {
+    fprintf(stderr, "total flushed: %" PRIu64 "\n", total_flushed);
+
     if (termw == -1) {
         fputs("tb_shutdown() should not be called twice.", stderr);
         abort();
@@ -186,6 +191,9 @@ void tb_present(void)
     }
     if (!IS_CURSOR_HIDDEN(cursor_x, cursor_y))
         write_cursor(cursor_x, cursor_y);
+
+
+    total_flushed += output_buffer.len;
     bytebuffer_flush(&output_buffer, inout);
 }
```

For your branch

```
diff --git a/Cargo.toml b/Cargo.toml
index 1d78da0..1e6cd08 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,5 +38,5 @@ chrono = "0.4.9"
 numtoa = "0.2.3"
 
 [profile.release]
-debug = false
 lto = true
+codegen-units = 1
diff --git a/termbox/src/lib.rs b/termbox/src/lib.rs
index 0456e61..2e514b7 100644
--- a/termbox/src/lib.rs
+++ b/termbox/src/lib.rs
@@ -28,6 +28,7 @@ pub struct Termbox {
     // (x, y)
     cursor: Option<(u16, u16)>,
     output_buffer: Vec<u8>,
+    total_flushed: u64,
 }
 
 struct CellBuf {
@@ -146,6 +147,7 @@ impl Termbox {
             last_bg: 0,
             cursor: Some((0, 0)),
             output_buffer: Vec::with_capacity(1024),
+            total_flushed: 0,
         };
 
         termbox.hide_cursor();
@@ -288,6 +290,9 @@ impl Termbox {
         }
 
         self.tty.write_all(&self.output_buffer).unwrap();
+
+        self.total_flushed += self.output_buffer.len() as u64;
+
         self.output_buffer.clear();
     }
 
@@ -457,5 +462,7 @@ impl Drop for Termbox {
         unsafe {
             libc::tcsetattr(libc::STDOUT_FILENO, libc::TCSAFLUSH, &self.old_term);
         }
+
+        eprintln!("Total flushed: {}", self.total_flushed);
     }
 }
```

emigr2k1:
Mhmm. I pad some zeroes every time I convert the cursor to string, and I think that's why it's 10x bytes. In the original branch it's only 4x more bytes. Maybe that and write! is the problem. I just don't know where the extra bytes come from.

osa1:
Ahh, I think I know why. When we show a character the terminal automatically increases the cursor's column, so we don't need to write "goto"s for consecutive characters in the same line. That's why the original termbox has `lastx` and `lasty` variables used, to keep track of the current cursor position. Update coming.

emigr2k1:
@osa1, Yea. Should I send a pr? Or will you work on it?

osa1:
I just confirmed that the redundant "goto"s are indeed the problem here. The branch `hack_test_2` has an implementation that is as fast as the C implementation -- for the benchmark that C code runs in 4.1 seconds `hack_test_2` runs in 4.2 seconds!

It'd be good to reduce the amount of unsafe code though. @emigr2k1 if you have the time it'd be good to start with the current branch (I just rebased it), and try these:

- Introduce cursor position caching to avoid redundant "goto" writes. This should make it significantly faster.
- Optimize termion's Goto so that the `Display` implementation doesn't go through `ToString` (which allocates an intermediate string).
- If still not enough, implement fast `goto` (as in `hack_test_2` branch).

When the performance is similar it'd be good to check the amount of flushed data, as a sanity check.

emigr2k1:
@osa1 sure! I don't have time to work on it right now, but in a pair of hours. So, if you start working on it before me, ping me to not duplicate work.

osa1:
OK, this is now as fast as the C implementation.

The code is a mess though. I'll do some refactoring, and document the invariants.

Since we already implemented the tricky bits from termion I think we should be able to remove the dependency with a little bit of work, I'll also try that.

Finally I'll implement some tests that test the wide char stuff. If terminals bump row number when something is printed to the last column then we can also reduce the `goto` calls even more. I'll test this too.

emigr2k1:
@osa1, I don't know if it was supposed to be fixed in the last commits you made, but printing wide chars is still a problem. For example, if you make a line of ten of this char:   駆   it's only going to print five. If you make a line like: 駆a駆a駆a駆a駆a it's going to skip the "a" letters.

osa1:
@emigr2k1 that's a known problem and I'm not sure how to fix it. Basically the TUI implementation (libtiny_tui) doesn't have a notion of wide character, so if you get a message e.g. `"駆a"` it shows `駆` in column `C` and then `q` in column `C+1` (instead of `C+2`) even though `駆` is two-characters wide. Then termbox when it renders column `C` it checks the character width, and because it's 2-characters wide it skips the next cell and doesn't render `a`.

I'm not sure if there's much to be done in termbox for this, I think this is a problem in libtiny_tui.

If you have any ideas on how to best fix this  I'd be happy to listen.

emigr2k1:
@osa1, I see. Sorry, I'm not familiar with the codebase so I'm not sure what would be a good way to fix it.
If I wanted to check on this, should I start with libtiny_tui then? So, I think this is a different issue and shouldn't be discussed here. Is there an existing issue for this?

osa1:
@emigr2k1 No need to apologize! It's great that you're interested in fixing this.

> If I wanted to check on this, should I start with libtiny_tui then?

I think that's a good idea, yes. The reason why libtiny_tui will need to be updated for this is because it splits long lines into multiple lines, handles resizing etc. which all need to be updated to take wide characters into account. For example if the screen is 5-characters wide and you have

    foo bar

libtiny_tui splits this into two lines and renders

    foo
    bar

But now suppose that you have this instead

    駆駆 駆 駆駆駆

Because libtiny_tui doesn't have a notion of wide character and assumes all characters take 1 column in terminal it splits this into

    駆駆 駆
    駆駆駆

(it only splits at a white space)

And we lose the last chracter in the first line (becuase first three characters take all the space).

> So, I think this is a different issue and shouldn't be discussed here.

Yeah. This is about rewriting the C code in Rust. It's fine if bugs/limitations in C code exists in the Rust version code too.

> Is there an existing issue for this?

No but it'd be great to have one. Screenshots would be even better.

PR #146, opened 2019-10-06
Optimize Goto's

emigr2k1:
This includes changes in the way Goto's are handled. First, it avoids
printing unnecesary Goto's. Second, it doesn't allocate when it needs to
print them.

This changes allow this termbox version to be as fast as the C version
of termbox.

### DIFF
--- a/termbox/src/lib.rs
+++ b/termbox/src/lib.rs
@@ -410,16 +410,11 @@ fn num_to_buf(mut num: u16, buf: &mut Vec<u8>) {
         }
     }
 
-    if chars_len == 1 {
-        return;
-    }
-
     let swap_len = start_len + (chars_len / 2) as usize;
-    let mut cur_char;
     let mut c = 0;
     for i in start_len..swap_len {
         let next_swap_idx = start_len + chars_len - c - 1;
-        cur_char = buf[next_swap_idx];
+        let cur_char = buf[next_swap_idx];
         buf[next_swap_idx] = buf[i];
         buf[i] = cur_char;
         c += 1;
--- a/termbox/src/lib.rs
+++ b/termbox/src/lib.rs
@@ -24,6 +24,7 @@ pub struct Termbox {
     last_bg: u16,
     // (x, y)
     cursor: Option<(u16, u16)>,
+    last_cursor: (u16, u16),
     output_buffer: Vec<u8>,
 }
 
@@ -142,6 +143,7 @@ impl Termbox {
             last_fg: 0,
             last_bg: 0,
             cursor: Some((0, 0)),
+            last_cursor: (0, 0),
             output_buffer: Vec::with_capacity(32 * 1024),
         };
 
@@ -216,6 +218,8 @@ impl Termbox {
     }
 
     pub fn present(&mut self) {
+        self.last_cursor = (0, 0);
+
         if self.buffer_size_change_request {
             self.update_size();
             self.buffer_size_change_request = false;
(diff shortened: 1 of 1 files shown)


termbox/src/lib.rs
```diff
@@ -405,6 +383,57 @@ impl Termbox {
...
+            write!(self.output_buffer, "{}", ch).unwrap();
+        } else {
+            goto(to_x, to_y, &mut self.output_buffer);
+            write!(self.output_buffer, "{}", ch).unwrap();
+        }
+        self.last_cursor = (to_x, to_y);
+    }
+}
+
+const LOOKUP: &[u8] = b"0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
osa1: This is only used to convert numbers 0-9 to the ASCII representations so this shouldn't have letters.

termbox/src/lib.rs
```diff
@@ -405,6 +383,57 @@ impl Termbox {
...
+        c += 1;
+    }
+}
+
+fn goto(x: u16, y: u16, buf: &mut Vec<u8>) {
+    write!(buf, "{}", "\x1B[").unwrap();
+    num_to_buf(y, buf);
+    write!(buf, "{}", ';').unwrap();
+    num_to_buf(x, buf);
+    write!(buf, "{}", 'H').unwrap();
```
osa1: Could you replace these `write!`s with `buf.extend_from_slice()`s? That's simpler (no magic macros) and you won't need to `unwrap()`.

termbox/src/lib.rs
```diff
@@ -216,6 +218,8 @@ impl Termbox {
     }
 
     pub fn present(&mut self) {
+        self.last_cursor = (0, 0);
```
osa1: Hmm, so I guess because `(0, 0)` is not a valid location (upper-left corner is `(1, 1)`) this effectively invalidates the last location. Did I get this right?

termbox/src/lib.rs
llogiq: Is the lookup needed at all, considering our value range is 0-9, and our output is `b'0'-`b'9'`?

termbox/src/lib.rs
```diff
@@ -405,6 +383,57 @@ impl Termbox {
...
+}
+
+const LOOKUP: &[u8] = b"0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+fn num_to_buf(mut num: u16, buf: &mut Vec<u8>) {
+    let start_len = buf.len();
+    let mut chars_len = 0;
+    loop {
+        let rem = num % 10;
+        let ch = LOOKUP[rem as usize];
```
llogiq: This could be `let ch = rem as u8 + b'0';`

osa1: Agreed, let's update this.

termbox/src/lib.rs
emigr2k1: Yes. The position is (1, 1) based.

termbox/src/lib.rs
emigr2k1: No. It is not needed.

termbox/src/lib.rs
emigr2k1: Sure. I suppose the other write! macros, for things like colors, are another issue?

termbox/src/lib.rs
emigr2k1: Agreed.

termbox/src/lib.rs
```diff
@@ -405,6 +383,55 @@ impl Termbox {
...
+        num /= 10;
+        chars_len += 1;
+        if num == 0 {
+            break;
+        }
+    }
+
+    if chars_len == 1 {
+        return;
+    }
```
osa1: This block can be removed.

termbox/src/lib.rs
```diff
@@ -405,6 +383,55 @@ impl Termbox {
...
+
+    let swap_len = start_len + (chars_len / 2) as usize;
+    let mut cur_char;
```
osa1: Can you remove this and just declare it in the next block? (`let cur_char = buf[next_swap_dx]`)

termbox/src/lib.rs
```diff
@@ -405,6 +383,55 @@ impl Termbox {
             .unwrap();
         }
     }
+
+    fn send_char(&mut self, to_x: u16, to_y: u16, ch: char) {
+        // if the target cell is next to the last cell, then don't move the cursor
+        if (self.last_cursor.0 + 1) == to_x && self.last_cursor.1 == to_y {
```
osa1: I just realized that this won't work as expected when the last printed char occupies more than one column. See the `cw` variable in `present()`.

This bug actually exists in the original termbox too..

Not sure what's the best way to fix this bug I guess we could maintain a `current_cursor` instead of `last_cursor`, and bump it after sending a character, using the character's visible width.

emigr2k1: What I'm seeing is that when the last character occupies 2 cells, the next character is not even sent to "send_char". It's being skipped.
I can't look at this right now. I'm in school so I'm just answering between classes. If you want to pick up this, or I can also check it in a few hours.

PR #147, opened 2019-10-07
Try both IPv4 and IPv6 addresses when domain resolves to both

### LINKED ISSUE #144: tiny fails to connect with "other os error"
I configured `tiny`, following the description in the `README.md`, but `tiny` fails to connect to any server (tried servers: `irc.mozilla.org, chat.freenode.net`) with the following message:
```
Connection error: other os error. Will try to reconnect in 30 seconds.
Disconnected. Will try to reconnect in 30 seconds.
```
I don't really know how to debug this.
I tried connecting with either TLS enabled or disabled, both don't work.
Possible related: my network has no outgoing IPv6 routes/internet acces, only IPv4, but uses IPv6 internally.

osa1:
See #144 for the problem this solves. When a domain name resolves to IPv4 and IPv6 addresses we now try to connect to both, concurrently, and continue with the one that connects first. This solves the problem of user only being able to connect to IPv4 or IPv6 addresses.

An alternative would be to ask the user to choose whether to use IPv6 or not, and default to IPv4. I think this version is better; it doesn't require any input from the user (no need for a new tinyrc field) -- it just works.

Fixes #144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Unreleased
+
+- When an IRC server domain resolves to IPv4 and IPv6 addresses tiny now selects
+  one IPv4 and one IPv6 address and tries to connect to both. This works better
+  when a user can only connect to IPv4 or on to IPv6, but address name resolver
+  returns e.g. IPv6 address first when the user can only connect to IPv4 (#144).
+
 # 2019/10/05: 0.5.0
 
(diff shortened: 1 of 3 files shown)


osa1:
Creating connections only to close right away is not great. I think we should just try all IP addresses that the address name resolver returns, sequentially.

osa1:
I think #148 is a better solution, closing this one.

PR #156, opened 2019-11-02
help command

eHammarstrom:
Should the other commands, such as `/clear`, move in alongside these?

### DIFF
--- a/libtiny_tui/src/messaging.rs
+++ b/libtiny_tui/src/messaging.rs
@@ -284,13 +284,22 @@ impl MessagingUI {
         self.msg_area.flush_line();
     }
 
+    fn _add_client_msg(&mut self, msg: &str) {
+        // Replace tabs with 4 spaces
+        let msg = msg.replace("\t", "    ");
+        let lines = msg.split('\n');
+        for line in lines {
+            self.msg_area.add_text(line);
+            self.msg_area.flush_line();
+        }
+    }
+
     pub(crate) fn add_client_err_msg(&mut self, msg: &str) {
         self.reset_activity_line();
 
         self.msg_area
             .set_style(SegStyle::SchemeStyle(SchemeStyle::ErrMsg));
-        self.msg_area.add_text(msg);
-        self.msg_area.flush_line();
+        self._add_client_msg(msg);
     }
 
     pub(crate) fn add_client_notify_msg(&mut self, msg: &str) {
@@ -298,8 +307,7 @@ impl MessagingUI {
 
         self.msg_area
             .set_style(SegStyle::SchemeStyle(SchemeStyle::Faded));
-        self.msg_area.add_text(msg);
-        self.msg_area.flush_line();
+        self._add_client_msg(msg);
         self.reset_activity_line();
     }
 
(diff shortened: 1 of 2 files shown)


osa1:
Thanks for doing this!

> Should the other commands, such as /clear, move in alongside these?

Yes! I think a new API method to the `UI` trait to get UI commands would be helpful. Maybe something like:

```rust
pub struct UICmd {
    pub name: String,
    pub description: String,
    pub usage_args: String,
}

pub trait UI {
    ...
    fn get_cmds(&self) -> Vec<UICmd>;
}
```

src/cmd.rs
```diff
@@ -20,12 +20,31 @@ pub(crate) struct Cmd {
     /// Command name. E.g. if this is `"cmd"`, `/cmd ...` will call this command.
     pub(crate) name: &'static str,
 
-    // Command help message. Shown in `/help`.
-    // pub(crate) help: &'static str,
+    /// Command description.
+    pub(crate) description: &'static str,
+
+    /// Command usage. Shown in `/help <command>` or when the command itself is misused.
+    pub(crate) usage_args: &'static str,
```
osa1: I think this should be a part of the description, for two reasons: (1) it saves users a command (2) currently it's not possible for a user to know that they need to run `/help cmd` to get argument info, they need to discover it somehow. If we show arguments in `/help` the UX will be slightly better IMO.

src/cmd.rs
```diff
@@ -110,9 +130,41 @@ static CMDS: [&Cmd; 8] = [
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+static HELP_CMD: Cmd = Cmd {
+    name: "help",
+    cmd_fn: help,
+    description: "List and lookup commands",
```
osa1: What does lookup mean here? I think "List commands" should be enough.

src/cmd.rs
eHammarstrom: What description format would you suggest?

This format can get quite long,
`command - description. Usage: /command args`

osa1: I don't find your examples too long, but maybe we can add a new line before the usage, something like:

```
away - Set and remove away status.
       Usage: /away [MESSAGE]
```

Feel free to suggest a different formatting.

I also wonder if we should give more details in the description. For example, `/away` unsets the away status, `/away <message>` sets it. Current description doesn't say this explicitly, and it's hard for a user to figure this out from this description.

libtiny_tui/src/messaging.rs
```diff
@@ -284,22 +284,30 @@ impl MessagingUI {
...
 
+    fn _add_client_msg(&mut self, msg: &str) {
+        // Replace tabs with 4 spaces
+        let msg = msg.replace("\t", "    ");
+        let lines = msg.split('\n');
+        for line in lines {
+            self.msg_area.add_text(line);
+            self.msg_area.flush_line();
+        }
+    }
```
osa1: Use 4 spaces instead of `\t` and you won't need this.

`\t` is really tricky to use with termbox so it's better if we use spaces instead.

eHammarstrom:
I am not going to add that is this patch.

The tab is translated to 4 spaces, tabs are not used. This is to save some space when writing the description and inserting spacing by using the tab control char.

osa1:
It seems like the author won't address my comments, closing.

PR #158, opened 2019-11-03
travis.yml: Use rustfmt to check code format

osa1:


osa1:
Seems to work!

This will probably break if there will be a nightly without a `rustfmt` component, but I think that's very rare (I update my toolchain regularly and never saw a nightly without rustfmt).

PR #163, opened 2019-12-24
Implement user configurable timestamp formatting

### LINKED ISSUE #159: Add option for higher resolution timestamp
Configuration setting for enabling `HH:MM:SS` for each irc message.

osa1:
Did this during a 11h flight today ..

Here's a screenshot of how this looks when I add `timestamp_format: '%A %H:%M:%S'` to my config file:

Screenshot_2019-12-24_07-46-25

This also re-enables the `/reload` command and makes a few other unrelated changes. I'll move those commit to the master branch and rebase this.

The code is not final, it has some bugs. I observed a crash while tweaking the format string and doing `/reload`. There also seems to be issues with splitting lines.

Fixes #159

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
   location. The old location `$HOME/.tinyrc.yml` is still used when there isn't
   a config file in the new location, to avoid breakage. `$HOME/.config` is used
   for `$XDG_CONFIG_HOME` when the env variable is not available (#152).
+- Timestamps can now be configured using the new (optional) `timestamp_format`
+  field in the config. Value should be a valid `strftime()` format string (see
+  `man strftime` for details). Default value is the same as before (`%H:%M`).
+  (#161)
 
 # 2019/10/05: 0.5.0
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ tiny is an IRC client written in Rust.
 
 - Configurable colors
 
+- Configurable timestamps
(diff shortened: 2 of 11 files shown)


osa1:
This is good to do. I'll test it for a bit before merging.

osa1:
OK so there's one problem with this approach which may or may not be a blocker. Suppose I have two lines added with `%H:%M` as the timestamp formatter, line 1 and line 2, and only line 1 has a timestamp because when formatted with `%H:%M` line 2 has the same timestamp as line 1. Now suppose I changed the formatter to `%H:%M:%S`. Now ideally both lines should have a timestamp, but currently we only update the existing timestamps (never add a new timestamp) so it looks as if the two lines are added at the same second even though the second component of their actual timestamps are different.

I'll have to think about how to fix this best.

osa1:
We should also switch to `time-0.2` as it has much more compact representation of times than 0.1.

osa1:
Here's the problem in action:

Screenshot_2019-12-27_07-38-15

I switched to a higher-resolution timestamp formatter to a lowe-resolution one which caused repeated timestamps (which we normally don't do).

osa1:
This is a simple feature but made very complicated because of the `/reload` command. I wonder if we should just remove `/reload` and provide users another way to test their TUI configs. The `Tm` structure is large and makes `Line` almost 10x larger. Code becomes much more complicated too...

PR #172, opened 2020-01-19
Optional rustls support

jbg:
Added a default feature `tls-native` with the status quo, and a non-default feature `tls-rustls` which uses rustls (with rustls-native-certs to use the platform's native cert store).

This also adds a dependency on lazy_static so that the TLS connector is only created once (saves parsing all the certs repeatedly), but lazy_static was already in the dependency tree due to several other packages in the tree depending on it.


osa1:
This looks great, thanks!

I don't know why Travis-CI doesn't show build errors next to commit lines now, but this needs a round of `rustfmt`. I'll do it myself after merging.

Opened an issue for the follow-up work: #172

PR #177, opened 2020-01-29
A framework for TUI tests

osa1:
This PR adds a way to test TUI rendering without actually doing any rendering to the terminal. An example test shows how to define a test:

```rust
#[test]
fn init_screen() {
    let mut tui = TUI::new_test(20, 10);
    tui.draw();

    #[rustfmt::skip]
    let screen =
        "|                    |
         |                    |
         |                    |
         |                    |
         |                    |
         |                    |
         |Any mentions to you |
         |will be listed here.|
         |                    |
         |mentions            |";

    expect_screen(screen, &tui);
}
```

This creates a new test instance of the TUI, draws it (not to the terminal, only to termbox's internal buffer), and then compares the screen with the given screen. The screen bounds are marked with `|`s.

There are pretty much no runtime overheads of this addition. Only changes in termbox is that it now works without a tty (some conditionals in a few places check if have a tty when drawing). When tty doesn't exist we only update the internal buffer. A new method `Termbox::get_front_buffer_str` returns a copy of the internal buffer as a `String`. We then compare that string with the provided specification in the tests. Simple enough!

One caveat is that we currently don't support color and style specifiers when generating the buffer string in `get_front_buffer_str`. That shouldn't be too hard to implement, I'm just not sure how to encode those in a `String`. One alternative would be to return a copy of the buffer itself, instead of encoding it in a string. That way the test framework would be free to check in different ways.

### DIFF
--- a/libtiny_tui/src/lib.rs
+++ b/libtiny_tui/src/lib.rs
@@ -18,6 +18,9 @@ mod tui;
 mod utils;
 mod widget;
 
+#[cfg(test)]
+mod tests;
+
 pub use crate::tab::TabStyle;
 pub use libtiny_ui::*;
 
--- a/libtiny_tui/src/tests/mod.rs
+++ b/libtiny_tui/src/tests/mod.rs
@@ -1,0 +1,107 @@
+use crate::tui::TUI;
+
+use termbox_simple::CellBuf;
+
+fn buffer_str(buf: &CellBuf, w: u16, h: u16) -> String {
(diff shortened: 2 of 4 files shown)


osa1:
I think it's a good idea to return a copy of the buffer directly from termbox, instead of encoding it in a string. I'll update.

osa1:
I actually found and fixed a bug while working on this. It turns out testing is useful!

PR #178, opened 2020-01-29
Make desktop notifications feature (dbus) optional

osa1:
With this I can make a statically linked executable with

    $ cargo build --target=x86_64-unknown-linux-musl \
        --no-default-features --features tls-rustls

The generated executable does not support generating desktop
notifications and uses rustls instead of libssl.

(#65)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,9 +9,6 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Install native dependencies
-        run: sudo apt-get install libdbus-1-dev pkg-config libssl-dev
-
       - name: Get nightly toolchain
         uses: actions-rs/toolchain@v1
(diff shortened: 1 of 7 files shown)


osa1:
Statically-linked tiny works great on Windows under WSL!

tiny

osa1:
I think I found a way to make dbus opt-out without making installation story worse: just make a `no-dbus` feature that is not optional. That way if a user wants to use rustls but not disable dbus they can still do `--no-default-fetaures --features tls-rustls`.

osa1:
Ugh.. I keep forgetting this. The idea above is not possible as cargo features are supposed to be "additive". I guess we'll have to write detailed installation instructions in README.

osa1:
I'm still not sure if this is a good idea, but I made rustls the default and desktop notifications opt-in. See commit msg for details.

PR #181, opened 2020-02-12
List cmd

flankstaek:
Added the /LIST command.

Let me know if you want any changes, first PR here, didn't see similar PR's or Contributing guidelines to follow so I just winged it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ Commands start with `/` character.
 
 - `/nick <nick>`: Change nick
 
+- `/list`: List all channels in current server
+
 - `/names`: List all nicks in the current channel. You can use `/names <nick>` to
   check if a specific nick is in the channel.
 
--- a/README.md
+++ b/README.md
@@ -178,8 +178,6 @@ Commands start with `/` character.
 
 - `/nick <nick>`: Change nick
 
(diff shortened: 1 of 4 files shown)


osa1:
Thanks for the patch.

I'm not sure if this is necessary as we kinda already have this feature: just send `LIST` in a server tab. Does this provide anything that sending `LIST` in a server tab does not?

flankstaek:
Nope, just adds a slash command for it as I'm not sure it's entirely clear (without looking at source) that sending LIST to a server would work, but don't think this is necessary to merge if you don't want to.

osa1:
> I'm not sure it's entirely clear (without looking at source) that sending LIST to a server would work

Right, we should mention README (maybe in "commands" section?) that sending a message in a server tab directly sends the message to the server, so it can be used to use parts of the IRC protocol that tiny doesn't support (like the `LIST` message). We could also give `LIST` as an example.

@flankstaek would you like to do that and update this PR? I think we don't need more code for this.

flankstaek:
Reverted the code changes and added a section to the README.

osa1:
Perfect, thanks @flankstaek !

Issue #186, opened 2020-04-13
Allow naming servers to give them a better display name


e00E:
Currently the address of the server is used to display the server tab. For some servers this can be unhelpful (ip address) or too long (very long domain name). I would like to be able to give my servers a helpful name in the config. This name would be used instead.

trevarj:
Implementing feature here:
https://github.com/osa1/tiny/pull/188

osa1:
This is now implemented. Just add an `alias` field to your server configuration. Example:

```
    - addr: chat.freenode.net
      alias: freenode
      ...
```

Here in the tab line you'll see `freenode` instead of `chat.freenode.net` as before.

e00E:
Thank you osa and trevarj.

PR #188, opened 2020-04-13
Allow server alias in config #186

trevarj:
Prototyping this...seems to work, but didn't fully test yet.

### DIFF
--- a/libtiny_ui/src/lib.rs
+++ b/libtiny_ui/src/lib.rs
@@ -91,7 +91,7 @@ pub trait UI: objekt::Clone {
     fn draw(&self);
 
     /// Create a new server tab.
-    fn new_server_tab(&self, serv: &str);
+    fn new_server_tab(&self, serv: &str, alias: Option<String>);
 
     /// Close a server tab and all channel and user tabs for that server.
     fn close_server_tab(&self, serv: &str);
@@ -177,9 +177,9 @@ impl<UI1: UI + Clone, UI2: UI + Clone> UI for CombinedUIs<UI1, UI2> {
         self.ui2.draw();
     }
 
-    fn new_server_tab(&self, serv: &str) {
-        self.ui1.new_server_tab(serv);
-        self.ui2.new_server_tab(serv);
+    fn new_server_tab(&self, serv: &str, alias: Option<String>) {
+        self.ui1.new_server_tab(serv, alias.clone());
+        self.ui2.new_server_tab(serv, alias);
     }
 
     fn close_server_tab(&self, serv: &str) {
--- a/tiny/src/config.rs
+++ b/tiny/src/config.rs
@@ -15,6 +15,10 @@ pub(crate) struct Server {
     /// Address of the server
     pub(crate) addr: String,
 
+    /// Optional server alias to be shown in the tab line.
+    #[serde(default)]
+    pub(crate) alias: Option<String>,
+
     /// Port of the server
     pub(crate) port: u16,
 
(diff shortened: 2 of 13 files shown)


tiny/src/main.rs
```diff
@@ -123,10 +123,12 @@ fn run(
         let mut clients: Vec<Client> = Vec::with_capacity(servers.len());
 
         for server in servers.iter().cloned() {
-            tui.new_server_tab(&server.addr);
+            let server_name = server.alias.clone().unwrap_or(server.addr.clone());
```
osa1, line 126: You shouldn't need cloning here as `new_server_tab` expects a ref.

trevarj: The clone is so that we can pass the values to `ServerInfo`  on line 130 and 131.

I can try it this way, if you want: 
```
            let server_name: String = match server.alias {
                Some(alias) => alias,
                None => server.addr.clone()
            };
            // let server_name = server.alias.unwrap_or(server.addr);
            tui.new_server_tab(&server_name);
```

osa1: Hmm, that's fair. Yet another possibility is

```
let server_name = server.alias.clone().unwrap_or_else(|| server.addr.clone());
```

This only clones `server.addr` if `server.alias` is `None`. However your code with `match` is best I think as it doesn't clone `alias` when it's not `None`. I suggest we go with the `match`.

osa1:
On a more careful reading I think this have a problem. Here's a screenshot with the problem:

Screenshot_2020-04-14_13-21-33

Note how in the same tab I'm seeing messages from two different servers. Similarly tabs for these two servers are interleaved. This happens when I use the same alias for multiple servers.

Currently TUI<->client communication is done using unique server names. This works fine when server names are actually domain names or IP addresses (currently the case) because we don't allow connecting to same server multiple times in a single tiny process. But when we allow giving names to servers we can't use "alias" as the unique server name as aliases are not unique.

What to do? Two options comes to mind:

- (easy) Make sure aliases are unique.
- (easy) Keep using server name as tab name, but when rendering use the alias.
- (harder) Give each connection unique numbers, use the numbers for TUI<->client communication. This also allows connecting to a server multiple times in a single tiny process.

I think I prefer (2).

trevarj:
I think I tried to implement method 2 originally but ran into issues when I realized the tab name is rendered from `MsgSource::visible_name`. I will look into it again.

osa1:
Thanks! It should be too hard. Let me know if you encounter any problems.

libtiny_ui/src/lib.rs
```diff
@@ -35,23 +35,29 @@ pub enum MsgSource {
...
     }
 
     pub fn visible_name(&self) -> &str {
         match *self {
-            MsgSource::Serv { ref serv, .. } => serv,
+            MsgSource::Serv {
+                ref serv,
+                ref alias,
+            } => match alias {
+                Some(alias) => alias,
```
trevarj, line 58: all of this for this one call!

osa1: Hmm.. I wonder if we could add a field to `Tab` instead of `MsgSource`, and update the `Tab::visible_name` method to use that field. We'd only use `MsgSource::visible_name` when the field is `None`. Does this make sense?

trevarj: This sounds way better actually. Will fix it.

tiny/src/config.rs
```diff
@@ -15,11 +15,13 @@ pub(crate) struct Server {
...
     pub(crate) addr: String,
 
+    #[serde(default)]
+    pub(crate) alias: Option<String>,
+
     /// Port of the server
     pub(crate) port: u16,
 
     /// Use tls
-    #[serde(default)]
```
trevarj, line 22: Was this fixed in your clippy warning commit?

osa1: This seems to be done in your commit.

osa1:
Thanks @trevarj !

Issue #185, opened 2020-04-12
Panic in pinger.rs


e00E:
I was editing a message using the text editor through the `C-x` hotkey. When I closed the editor I saw this message interleaved with the channel text (probably because it went to stderr). Despite the panic tiny did not actually crash. It displayed the normal ping timeout message and successfully reconnected afterwards.

> panicked at 'called `Result::unwrap()` on an `Err` value: Closed(SendPing)', libtiny_client/src/pinger.rs:40:9

https://github.com/osa1/tiny/blob/f8477c322a0eb464072542f75c50ecfeb587bbd0/libtiny_client/src/pinger.rs#L40

osa1:
Thanks for reporting. I think this is just a manifestation of a long-standing problem with the `C-x` implementation. It currently stops *all* tasks, that includes the tasks that drive connections, timer tasks for checking connectivity etc. If you spend long-enough time in your editor after `C-x` it causes all kinds of problem.

The solution should be to run the editor in a new thread, allowing background tasks to run. This may be as easy as wrapping the call with `tokio::task::spawn_blocking`. Only difficulty with this is that the TUI task needs to wait for the blocking task to return.

osa1:
I thought a little bit about how to best fix this isssue with `C-x`. Here's an idea:

- First of all `termbox` should support `present()` when suspended. It'd basically just update the internal buffer but do not write the buffer to the tty when suspended, until it's activated again.
  This is easy to implement as `termbox` already works without a tty (used in tests).

- Once we have that we'll be able to suspend the TUI while still being able to do updates. This will allow us to show `$EDITOR` in a new thread while all the connection tasks update the TUI.

- TUI is not `Send` or `Sync` so we won't be able to have a reference to it in the `$EDITOR` thread. Instead we will send an event to the TUI via the event channel. The problem here is that normally TUI only has the receiver end of this channel. So to be able to do it we'll also keep a sender in TUI, and clone and pass it to the `$EDITOR` thread. This will also require a new event type.

After this it should be possible to run `$EDITOR` for hours if we wanted.

PR #195, opened 2020-05-11
Avoid blocking tokio event loop in C-x

osa1:
This is an attempt at fixing #185.

Currently stuck. The problem is after the new `editor::edit` async function we have to transitively turn use sites into async function, which means UI methods will be async now. Not sure if this is a good idea.

The first commit is a useful refactor and improves UX around this feature a little bit, so it should be merged regardless.

I wonder if we could make text_field allow multi-line editing instead.


osa1:
So I found a way to implement this without running TUI in its own thread (not task!), but it has another problem: the TUI task and editor thread races to read `stdin`.

We need to somehow stop TUI from reading `stdin` when an editor is running.

osa1:
Fixed in #222

PR #198, opened 2020-05-17
Replace term_input with crossterm

osa1:
Problems:

- We lose the `String` event provided by term_input when large amount of characters available in the input buffer.
- ~~crossterm doesn't generate `alt-right`/`alt-left` keys so we can't use those keys for moving tabs.~~ Fixed in my fork.


osa1:
I forked crossterm and added alt-arrow support.

osa1:
crossterm PR: https://github.com/crossterm-rs/crossterm/pull/431

osa1:
> We lose the String event provided by term_input when large amount of characters available in the input buffer.

I think we'll have to live with this.

osa1:
I just realized that crossterm spawns an OS thread when `EventStream` is used: https://github.com/crossterm-rs/crossterm/blob/master/src/event/stream.rs#L99-L111 not sure what that is necessary, but it'd be good to avoid it.

osa1:
Given that we lose a functionality with this and the extra runtime overhead (a thread), I think this may not worth it. I've improved `term_input` in #199 and added a missing key combination.

PR #199, opened 2020-05-20
term_input: A more efficient key combination parser

osa1:
Instead of linearly searching through a few dozen of items when a key
combination is pressed, we now generate a more efficient parser with
help of a proc macro for the combination. The generated function looks
at every byte in the input once. As an example of how this works, for
this declaration:

    byte_seq_parser! {
        example -> Key,

        [1, 2] => Key::Del,
        [1, 3] => Key::Home,
        [0] => Key::PageUp,
    }

This is the generated function

    fn example(buf: &[u8]) -> Option<(Key, usize)> {
        let mut i: usize = 0;
        match buf.get(i) {
            None => None,
            Some(byte) => match byte {
                1u8 => {
                    i += 1;
                    match buf.get(i) {
                        None => None,
                        Some(byte) => match byte {
                            3u8 => {
                                i += 1;
                                Some((Key::Home, i))
                            }
                            2u8 => {
                                i += 1;
                                Some((Key::Del, i))
                            }
                            _ => None,
                        },
                    }
                }
                0u8 => {
                    i += 1;
                    Some((Key::PageUp, i))
                }
                _ => None,
            },
        }
    }

(This way of matching a set of strings probably has a name, but I don't
know what it is!)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -504,7 +504,7 @@ dependencies = [
  "serde",
  "serde_yaml",
  "tempfile",
- "term_input 0.1.5",
+ "term_input 0.2.0",
  "termbox_simple",
  "time",
  "tokio",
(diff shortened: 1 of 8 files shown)


osa1:
Not sure why Github is not running the CI jobs?

PR #201, opened 2020-05-26
Implemented scroll wheel to scroll message area. Closes #130

### LINKED ISSUE #130: Implement scrolling with mouse wheel

trevarj:
Simple addition. Also really nice to use, in my opinion. Let me know if I missed something! Thanks.

### DIFF
--- a/libtiny_tui/src/messaging.rs
+++ b/libtiny_tui/src/messaging.rs
@@ -1,4 +1,4 @@
-use term_input::Key;
+use term_input::{Key, MouseWheel};
 use termbox_simple::Termbox;
 
 use std::convert::From;
@@ -163,12 +163,12 @@ impl MessagingUI {
                 WidgetRet::KeyHandled
             }
 
-            Key::ShiftUp => {
+            Key::ShiftUp | Key::MouseButton(MouseWheel::WheelUp) => {
                 self.msg_area.scroll_up();
                 WidgetRet::KeyHandled
             }
 
-            Key::ShiftDown => {
+            Key::ShiftDown | Key::MouseButton(MouseWheel::WheelDown) => {
                 self.msg_area.scroll_down();
                 WidgetRet::KeyHandled
             }
--- a/term_input/src/lib.rs
+++ b/term_input/src/lib.rs
@@ -40,6 +40,7 @@ pub enum Key {
     ShiftDown,
     ShiftUp,
     Tab,
+    MouseButton(MouseWheel),
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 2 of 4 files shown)


osa1:
Thanks! One problem though: I can't select text anymore in my terminal (st). Is that the case in your terminal as well? For example, someone just pasted a link in a channel, and I can't select & copy the link anymore.

trevarj:
Welp! 😭 Yeah, that doesn't work for me either. Too good to be true!

osa1:
If you look at the sequence to enable these events:

```
\x1b[?1000h\x1b[?1002h\x1b[?1015h\x1b[?1006h
```

this seems to be doing 4 things -- there are 4 occurrences of `\x1b[?`. I'm wondering why we have 4 such sequences, perhaps only one or two of them are for mouse wheel events and we can get away with only enabling those.

osa1:
I just checked [zulip-terminal][1] for how they handle mouse events: normally all events are interpreted by the application (including mouse clicks and selection), but if you press `shift` then the mouse control is given back to the terminal (until you release `shift`).

It's not ideal for tiny though: I'd rather only get mouse wheel events, and leave selection to the terminal always.

[1]: https://github.com/zulip/zulip-terminal

trevarj:
Sadly, those are xterm and rxvt escape codes to turn on/off mouse capture. Nothing specific to which buttons.

xterm ones can be found here: https://www.xfree86.org/4.8.0/ctlseqs.html#Mouse%20Tracking

```
 CSI ? P m h      DEC Private Mode Set (DECSET)
 
                  P s = 1 0 0 0 → Send Mouse X & Y on button press and release. See the section Mouse Tracking.
                  P s = 1 0 0 1 → Use Hilite Mouse Tracking.
                  P s = 1 0 0 2 → Use Cell Motion Mouse Tracking.
```

trevarj:
Thoughts on implementing highlighting and saving to the clipboard (cross-platform `clipboard` crate)?

osa1:
I don't think this is worth adding more dependencies and code to maintain. We already have key bindings `shift-arrow` and page-up/page-down for scrolling, this is just for convenience. If there isn't an easy and low-maintenance way of implementing this then I think we can live with the existing key bindings.

trevarj:
Understandable. 
The only thing I can think of is to have the first scroll on the mouse wheel toggle the mouse mode, and then when you don't scroll for a while it toggles it back off. Seems a little messy though.

trevarj:
On my term `Alt+Shift` allows me to highlight then `Ctrl-c` to copy.

osa1:
Yeah that works for me too, but not teaching every user how to select and copy is easier, and provides good UX. You do whatever you've been doing already and it works! Best UX.

Issue #169, opened 2020-01-01
aligned table display


eoli3n:
Is this possible to be able to 
- enable clock at each line in conf file
- enable align and specified nb of char for nick place as `` align: 15 ``

**from**
```                                                                                                
19:29 ahub: c'est quoi ton client eoli3n__ ?                                                                          
19:30 eoli3n__: https://github.com/osa1/tiny || tiny-irc-client-git AUR                                               
eoli3n__: je viens de l'installer                                                                                     
ahub: connais pas                                                                                                     
ahub: c'est quoi la différence avec un weechat ou autre ?                                                             
eoli3n__: je viens de decouvrir c'est top                                                                             
eoli3n__: ahub un simple fichier de conf, pas besoin de plugin pour gérer les highlight et les notif                  
ahub: ho                                                                                                             
19:31 ahub: ok :)                                                                                                     
19:32 -ahub                                                                                         
eoli3n__:    
```
**to**
```                                                                                                
19:29 ahub:        c'est quoi ton client eoli3n__ ?                                                                          
19:30 eoli3n__:    https://github.com/osa1/tiny || tiny-irc-client-git AUR                                               
19:30 eoli3n__:    je viens de l'installer                                                                                     
19:30 ahub:        connais pas                                                                                                     
19:30 ahub:        c'est quoi la différence avec un weechat ou autre ?                                                             
19:30 eoli3n__:    je viens de decouvrir c'est top                                                                             
19:30 eoli3n__:    ahub un simple fichier de conf, pas besoin de plugin pour gérer les highlight et les notif                  
19:31 ahub:        ho                                                                                                             
19:31 ahub:        ok :)                                                                                                     
19:32 -ahub                                                                                         
eoli3n__:    
```

trevarj:
Another formatting idea. Perhaps it isn't as compact and "tiny" as it should be though...
```
19:29 |      ahub: c'est quoi ton client eoli3n__ ?                                                                          
19:30 |  eoli3n__: https://github.com/osa1/tiny || tiny-irc-client-git AUR                                               
19:30 |  eoli3n__: je viens de l'installer                                                                                     
19:30 |      ahub: connais pas                                                                                                     
19:30 |      ahub: c'est quoi la différence avec un weechat ou autre ?                                                             
19:30 |  eoli3n__: je viens de decouvrir c'est top                                                                             
19:30 |  eoli3n__: ahub un simple fichier de conf, pas besoin de plugin pour gérer les highlight et les
[...]

PR #207, opened 2020-06-06
Align nicks and messages

ii8:
Personally I prefer when nicks and wrapped messages are aligned, merge it if you agree I guess.
Otherwise it could also be made configurable quite easily.

Review carefully though, first time using rust so there are likely mistakes.

### DIFF
--- a/libtiny_tui/src/messaging.rs
+++ b/libtiny_tui/src/messaging.rs
@@ -266,11 +266,15 @@ impl MessagingUI {
 ////////////////////////////////////////////////////////////////////////////////
 // Adding new messages
 
+const TS_WIDTH: i32 = 6;
+
 impl MessagingUI {
     fn add_timestamp(&mut self, ts: Timestamp) {
         if let Some(ts_) = self.last_activity_ts {
             if ts_ != ts {
                 ts.stamp(&mut self.msg_area);
+            } else {
+                self.msg_area.add_text("      ");
             }
         } else {
             ts.stamp(&mut self.msg_area);
@@ -283,6 +287,7 @@ impl MessagingUI {
 
         self.msg_area
             .set_style(SegStyle::SchemeStyle(SchemeStyle::Topic));
+        self.msg_area.set_indent(TS_WIDTH);
         self.msg_area.add_text(topic);
 
         self.msg_area.flush_line();
@@ -327,11 +332,13 @@ impl MessagingUI {
     ) {
         self.reset_activity_line();
         self.add_timestamp(ts);
+        let mut indent: i32 = TS_WIDTH;
 
         if is_action {
             self.msg_area
                 .set_style(SegStyle::SchemeStyle(SchemeStyle::UserMsg));
             self.msg_area.add_text("** ");
+            indent += 3;
         }
 
         {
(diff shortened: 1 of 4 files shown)


Kabouik:
> merge it if you agree I guess

Just chiming in to say that I will use that if it is merged or made an option.

trevarj:
We were having a discussion about this here: https://github.com/osa1/tiny/issues/169#issuecomment-614451588

ii8:
Yes, I saw that discussion after having already made this PR(should have checked issues first :facepalm:)

I'm just going to leave this PR lying here until a decision is made.

osa1:
Thanks for the PR. tiny's TUI is supposed to be compact and simple, but these kind of alternative layout requests came a few times in the past, and I don't want reject feature requests if they can be made optional without too much effort and in a way that doesn't make the default harder to maintain or test, or buggy, or slower.

So far I've seen these requested:

- Align after timestamps (this PR)
- Align after nicks (the issue linked above)

There are also other related requests like:

- Make it possible to configure timestamp formatting (#159)
- Make it possible to show timestamps always (I can't find the issue now, but I'm pretty sure this was requested in the past)

These are all related. For example, if we make timestamp formatting configurable we will have to remove `TS_WIDTH` in this PR (as the timestamp length won't be fixed anymore). So perhaps we should take the time to design and implement all these at once, as I'm pretty sure people will request these later again, and if we only do some of these we'll have to revert/update some of the changes in the previous PRs.

I'll try to see if there's an easy way to update the TUI to handle all these cases (align after timestamps, align after nicks, with configurable timestamp formatting), but if you've already looked at this ideas are always welcome.

ii8:
For different length timestamps TS_WIDTH will just have to become a variable, no big upset. Showing timestamps always is just one more config option and if-branch.
This PR does align messages that wrap with the beginning of that message text. It does not align the ends of all nicks, but I'm not sure that's even desirable(Everything would move when somebody with a longer nick starts talking, long nicks would cause a lot of empty space etc.)

However if you want complete customizability, adding extra characters like vertical lines to separate timestamps and nicks, timestamps on the far right of the screen instead of the left etc. then things would probably have to be changed more radically.

ii8:
Among those the only tricky one is aligning the end of nicks. It's not obvious how to deal with long nicks.

Kabouik:
I realize with this discussion that the only thing I am missing currently is keeping nicknames left-aligned when no timestamp is shown. Apart from that, I like that timestamps are shown only when necessary, or that Tiny uses all space available and does not right-align nicknames, or wraps multi-line messages below the nickname of the poster (but it shouldn't wrap below the timestamps if nicks are left-aligned).

meain:
I had tried doing this a while back. I used it for a while and switched back.

I don't think this should be the default though as the current config of tiny actually has a lot of usable space even on small screens.

About the max-length, I think there is limit of 9 chars for a nick. Not sure if all the implementations respect it though

osa1:
This feature was implemented by @trevarj a while ago. Not officially announced in README and CHANGELOG yet (few bugs left to fix before that), but you can enable it in master branch by adding `layout: aligned` to your config. Closing.

PR #209, opened 2020-06-12
Implemented DCC get for downloading files.

### LINKED ISSUE #206: [Feature request] Support for DCC file transfers
Some IRC servers with search bots are based on DCC transfers to send search results in the .txt format to the user, such as #bookz or #ebooks and others. 

`tiny` sends a notification about the DCC transfer being initiated, but can't actually receive anything. I am afraid it would be a lot of work to implement, but it is worth asking. For reference, this is how `irssi` handles DCC transfers: https://irssi.org/documentation/help/dcc/

trevarj:
Does not yet support sending files.
Closes #206


trevarj:
There's only one server that I know that uses this and it's the one referenced in #206.

Basically, you just go on there and send a command to search. 
The bot messages you privately and sends you a DCC SEND, which Tiny will pick up and store in the DCC Cache. It also helpfully prints out the command to use to download the file. 
Once the download is complete you get a notice in the shared server tab between you and the sender (in this case Irchighway's tab).

The download directory is configurable in `config.yml`.

image

libtiny_client/src/lib.rs
```diff
@@ -660,3 +695,127 @@ async fn try_connect(
...
+    tokio::task::spawn_local(async move {
+        let stream = match timeout(
+            tokio::time::Duration::from_secs(25),
+            tokio::net::TcpStream::connect(addr),
+        )
+        .await
+        {
+            Err(err) => {
+                error!("{}", err);
+                let err_msg = "File transfer connection timed out.";
```
trevarj, line 719: A strange issue is happening here where the error message gets truncated to "File". I was wondering if it has to do with the mpsc buffer, so I changed `try_send` to `send` but the same thing happened. It **does** send the correct message when the file transfer is successful.

Kabouik:
Are there any blockers that prevent merging this?

osa1:
@trevarj let me know if this is ready (as in it works and is complete) and has no unrelated changes, I can review this after #344.

As discussed before on IRC, a PR description with summary of the changes (both high level and in implementation) would be helpful.

trevarj:
@osa1 this one actually might be farther along than #344. I updated the description and added a gif so you can see how it works.

travankor:
Could this be made optional for those that want to ignore DCC CTCP?

trevarj:
> Could this be made optional for those that want to ignore DCC CTCP?

I can feature-gate this, if that makes more sense. @osa1 can make the final decision

Kabouik:
Has this PR been abandoned? Still using Tiny daily here a after all these years, even though I haven't updated in a while. I still love its compact layout.

Issue #202, opened 2020-05-30
Long, spaceless messages get cut off


trevarj:
Long messages that don't have enough spaces to line wrap get truncated.

I believe this is due to the `draw()` in `line.rs` and how it handles when there is no whitespace to split on.
```
                    // Not possible to split. Need to make sure we don't render out
                    // of bounds.
                    if col - pos_x < width {
                        if line >= first_line {
                            tb.change_cell(col, pos_y + line, char, sty.fg, sty.bg);
                        }
                        col += 1;
                    }
```

This isn't a huge problem. Personally, I would just resize the window and hope that the message/link can all fit, but it's probably a pain for people using a tiling window manager...

Wide view:
![image](https://user-images.githubusercontent.com/5448324/83335246-26622280-a2b4-11ea-9a34-e0f347a91436.png)
Compact view:
![image](https://user-images.githubusercontent.com/5448324/83335256-337f1180-a2b4-11ea-87d7-71c298c95373.png)

Tested on master

osa1:
Thanks for reporting. I thought I handled this, but apparently not. I think this is pretty bad as it can be misleading. For example, if someone pasted a long link and the window width is small, the link will cut off and when you copy you'll only copy the prefix.

We should split the line at the last visible character when this happens.

trevarj:
> We should split the line at the last visible character when this happens. 

I'm afraid this will be a messy add on to how it's being done now. Similar to the text input problem with calculating the number of lines...

Issue #208, opened 2020-06-12
Input area improvements/tasks


osa1:
I was reading the code today and found some potential improvements and a problem in the code.

- [ ] Starting with the problem: `InputArea::height` is never set. It starts its life as `None` and stays that way. That means `InputArea::get_height` always recomputes the height even if nothing changes in `InputArea`.

- [ ] I think it'd be better to invalidate `InputLine::ws_indices` (renamed from `InputLine::line_starts`) on update. Currently we don't update it until `calculate_height` is called, and if I update a line and forget to call `calculate_height` then `draw` will render input area incorrectly.

- [ ] We should have a fast path for the common case of inserting at the end. Currently this causes exponential behavior on paste, where we turn a long string into a series of key press events (see `handle_input_event`), and for every key press we compute the whitespace indices from scratch. I think `InputLine::insert` could be easily updated to handle this and efficiently update `ws_indices` when inserting at the end.

  This also means `InputLine::calculate_height` shouldn't generate `ws_indices` from scratch if it's already up-to-date (related to the second point above), otherwise the update in `insert` will never be used.

cc @trevarj

trevarj:
Thanks for taking a look and fixing it up. I read through your changes and they were really helpful for my learning.

osa1:
I edited my original post: pasting causes "quadratic" behavior, not "exponential"!

trevarj:
1. This first issue seems like it may be as as simple as putting `self.height = Some(line_count);` at the end of `InputArea::calculate_lines()`. I tested it a little, but not much.

2. Invalidating `ws_indices` seems like a good idea. When would you update and forget to call calculate height though? Does this just mean if the code were to change?

3. Should we also re-write the paste functionality to do a memcpy into the line buffer instead, then invalidate `ws_indices` and `InputArea::height`?

osa1:
Copying my comment from #210:

> Btw, I realized that my remark about quadratic behavior on paste in #208 is incorrect: we call `TUI::keypressed` for every char in the pasted string, but `keypressed` doesn't draw after handling the key. So we call `draw` just once after handling all the keypresses. Sorry for the false alarm!

osa1:
#210 fixed the issues here, but I just added one more relevant item which is for efficiently removing from the end.

PR #210, opened 2020-06-16
Caching InputLine calculation. Caching InputArea::height.

trevarj:
First attempt at solving some of the issues on #208

### DIFF
--- a/libtiny_tui/src/input_area/input_line.rs
+++ b/libtiny_tui/src/input_area/input_line.rs
@@ -2,22 +2,168 @@ use crate::{config::Colors, termbox, utils};
 use std::{cmp::min, ops::RangeBounds, vec::Drain};
 use termbox_simple::Termbox;
 
+/// Cache that stores the state of InputLine's height calculation.
+/// `line_count` is used as the dirty bit to invalidate the cache.
+#[derive(Clone, Debug)]
+struct LineDataCache {
+    /// Indices to split on when we draw, not always whitespaces
+    split_indices: Vec<i32>,
+    /// The total number of lines (height) that will be rendered
+    line_count: Option<i32>,
+    /// The current width of InputArea. Used in determining if we need to invalidate due to resize.
+    width: i32,
+    /// The current width of the InputLine (may be shorter due to nickname)
+    line_width: i32,
+    /// Current nickname length. Used in determining if we need to invalidate due to resize.
+    nick_length: usize,
+    /// The index into InputLine::buffer of the last whitespace that we saw in calculate_height()
+    last_whitespace_idx: Option<i32>,
+    /// True if the last character was a whitespace character
+    prev_char_is_whitespace: bool,
+    /// The length of the current line that is being added to.
+    /// Used to determine when to wrap to the next line in calculate_height()
+    current_line_length: i32,
+}
+
+impl LineDataCache {
+    fn new() -> LineDataCache {
+        LineDataCache {
+            split_indices: Vec::new(),
+            line_count: None,
+            width: 0,
+            line_width: 0,
+            nick_length: 0,
+            last_whitespace_idx: None,
+            prev_char_is_whitespace: false,
+            current_line_length: 0,
+        }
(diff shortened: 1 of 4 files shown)


osa1:
Thanks for doing this!

I did a quick review. Questions/comments:

- Do we really need all these fields in `LineDataCache`? For example, as far as I understand it doesn't make sense to re-use `last_whitespace_idx` and `prev_char_is_whitespace` from a previous call to `calculate_height`, so those states should be local to `calcualte_height`.

- Related to the point above: if you document these fields for what they are/ how they are used/ why they are needed the it may become clear which ones are needed in this type and which ones could be moved to the function `calculate_height` as a local variable.

- Methods of `LineDataCache` don't need to be `pub`.

- If I add a character to the end of the buffer we still run the whole computation from scratch, right? I think we could have a special method for inserting at the end that runs much faster. 

Btw, I realized that my remark about quadratic behavior on paste in #208 is incorrect: we call `TUI::keypressed` for every char in the pasted string, but `keypressed` doesn't draw after handling the key. So we call `draw` just once after handling all the keypresses. Sorry for the false alarm!

trevarj:
> Do we really need all these fields in LineDataCache? For example, as far as I understand it doesn't make sense to re-use last_whitespace_idx and prev_char_is_whitespace from a previous call to calculate_height, so those states should be local to calcualte_height.

I do believe we need those variables in order to save the state of the calculation, due to the "fast track" scenario, where we insert into the end of the buffer. It allows for resuming the calculation at the point where we left off.

> Related to the point above: if you document these fields for what they are/ how they are used/ why they are needed the it may become clear which ones are needed in this type and which ones could be moved to the function calculate_height as a local variable.

I can do this.

> Methods of LineDataCache don't need to be pub.

Done.

> If I add a character to the end of the buffer we still run the whole computation from scratch, right? I think we could have a special method for inserting at the end that runs much faster.

Same as explanation above, but here is the change to insert()
```
    /// Interface for Vec::insert()
    pub(crate) fn insert(&mut self, idx: usize, element: char) {
        self.buffer.insert(idx, element);
        if idx == self.buffer.len() - 1 {
            self.line_data.calculate_height(&self.buffer, idx);
        } else {
            self.line_data.set_dirty();
        }
    }
```
Note that when the idx is the end of the buffer, we resume the calculation from where it ended before. The call to `calculate_height` with an `offset` of `idx` will look at the state and decide if an index needs to be added to `ws_indices`. Maybe the `offset` idea isn't so great, and it could just be a boolean like `append` or something.

trevarj:
Rebased with master now build fails

libtiny_tui/src/input_area/input_line.rs
```diff
@@ -2,30 +2,175 @@ use crate::{config::Colors, termbox, utils};
 use std::{cmp::min, ops::RangeBounds, vec::Drain};
 use termbox_simple::Termbox;
 
+/// Cache that stores the state of InputLine's height calculation.
+/// `line_count` is used as the dirty bit to invalidate the cache.
+#[derive(Clone, Debug)]
+struct LineDataCache {
+    /// Indices to split on when we draw, not always whitespaces
+    ws_indices: Vec<i32>,
```
osa1, line 10: Ahh, so you also fix #202 with this, right? Let's rename this to `split_indices`.

trevarj: No, it doesn't fix that #202, sadly. That issue existed before any of the input changes. I've been thinking about how to fix that problem though.

I will rename.

osa1: Ah, I'm completely confused. This is about the input field of course it doesn't fix #202.

osa1:
Thanks @trevarj , this looks good to me. I tested this a little bit and I can see that caching works, and we update the state after the insertion point (instead of creating it from scratch) on key presses and also when pasting.

We could potentially add one more special case for deleting from the end (in `InputLine::remove`), but this is fine for now.

PR #216, opened 2020-06-22
Refactored line.rs height calculation and drawing.

### LINKED ISSUE #202: Long, spaceless messages get cut off

trevarj:
Reused LineDataCache. Basically a drop-in fix. 
I did have to manually go through each of the failing tests, because they weren't accurate to how text should get rendered.

Closes #202

- [x] Don't render whitespaces that are split on?

### DIFF
--- a/libtiny_tui/src/line_split.rs
+++ b/libtiny_tui/src/line_split.rs
@@ -1,0 +1,152 @@
+/// Cache that stores the state of a line's height calculation.
+/// `line_count` is used as the dirty bit to invalidate the cache.
+#[derive(Clone, Debug)]
+pub struct LineDataCache {
+    /// Indices to split on when we draw, not always whitespaces
+    split_indices: Vec<i32>,
+    /// The total number of lines (height) that will be rendered
+    line_count: Option<i32>,
+    /// The current width of InputArea. Used in determining if we need to invalidate due to resize.
+    width: i32,
+    /// The current width of the InputLine (may be shorter due to nickname)
+    line_width: i32,
+    /// Current nickname length. Used in determining if we need to invalidate due to resize.
+    nick_length: usize,
+    /// The index into InputLine::buffer of the last whitespace that we saw in calculate_height()
+    last_whitespace_idx: Option<i32>,
+    /// True if the last character was a whitespace character
+    prev_char_is_whitespace: bool,
+    /// The length of the current line that is being added to.
+    /// Used to determine when to wrap to the next line in calculate_height()
+    current_line_length: i32,
+    /// If the line of text has a cursor
+    has_cursor: bool,
+}
+
+impl LineDataCache {
+    pub fn new(has_cursor: bool) -> LineDataCache {
+        LineDataCache {
+            split_indices: Vec::new(),
+            line_count: None,
+            width: 0,
+            line_width: 0,
+            nick_length: 0,
+            last_whitespace_idx: None,
+            prev_char_is_whitespace: false,
+            current_line_length: 0,
+            has_cursor,
+        }
+    }
+
+    /// Performs a check to see if the width or nickname length changed
+    /// which would require an invalidation of the cache and recalculation of
+    /// the line height.
+    pub fn needs_resize(&self, width: i32, nick_length: usize) -> bool {
+        self.width != width || self.nick_length != nick_length
+    }
+
+    /// Sets `line_count` to `None`, which invalidates the cache.
+    pub fn set_dirty(&mut self) {
+        self.line_count = None;
+    }
+
+    /// Checks if the cache is invalidated by seeing if
+    /// `line_count` is `None`.
+    pub fn is_dirty(&self) -> bool {
+        self.line_count.is_none()
+    }
+
+    /// Resets the cache to a default state that requires
+    /// a height calculation.
+    pub fn reset(&mut self, width: i32, nick_length: usize) {
+        self.split_indices.clear();
+        self.line_count = None;
+        self.width = width;
+        self.nick_length = nick_length;
+        self.line_width = width - nick_length as i32;
+        self.last_whitespace_idx = None;
+        self.prev_char_is_whitespace = false;
+        self.current_line_length = 0;
+    }
+
+    pub fn get_line_count(&self) -> Option<usize> {
+        self.line_count.map(|c| c as usize)
+    }
+
+    pub fn get_splits(&self) -> &[i32] {
+        &self.split_indices
+    }
+
+    /// Function that calculates the height of the line.
+    /// and sets `split_indices` for drawing.
+    /// An `offset` allows for resuming the calculation - see InputLine::insert().
+    /// `offset` must be less than or equal to the current buffer size.
+    ///
+    /// Scans through the buffer in one pass to determine how many lines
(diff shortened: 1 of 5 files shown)


osa1:
Hmm.. I wonder if we could get rid of the `Seg` type entirely and work on `String`s. That'd reduce the memory usage even more, and looking at the code now, I don't see why we couldn't use `String` instead of `Vec<Seg>` in `Line`.

trevarj:
> I don't see why we couldn't use String instead of Vec<Seg> in Line.

Ok, I will do this.

trevarj:
@osa1 I used a smaller file than you, but I actually saw a better result with this new change than what's on master. I did verify with bench.rs that the previous commit was inefficient . Let me know if you can send me that massive text file, or try it out on your PC.

libtiny_tui/src/msg_area/line.rs
```diff
@@ -85,35 +80,48 @@ impl Seg {
...
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl DerefMut for StyledString {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.inner
+    }
+}
```
osa1, line 94: This is considered an anti-pattern by the community: https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deref.md we also consider this an anti-pattern in the code base at work. I don't think not having this will cause a lot of pain, doing `self.inner.foo` instead of `self.foo` and it's easier on the reader, so let's remove these.

libtiny_tui/src/msg_area/line.rs
```diff
@@ -1,34 +1,29 @@
...
-    /// substrings.
-    splits: Vec<i32>,
+    line_data: LineDataCache,
 }
 
 #[derive(Debug)]
-struct Seg {
-    text: String,
+struct StyledString {
+    inner: String,
```
osa1, line 26: IME the `inner` naming convention is mostly used when you wrap a struct with reference types and give it a new name, e.g. we have

```rust
pub struct TUI {
    inner: Weak<RefCell<tui::TUI>>,
}
```

For this type I think it's better to just call it `str` or `string`.

libtiny_tui/src/line_split.rs
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+
+        // Last line length is `line_width`, make room for cursor
+        if self.has_cursor && self.current_line_length == self.line_width {
+            temp_count += 1;
+        }
+        self.line_count = Some(temp_count);
+    }
+}
```
osa1, line 152: This is moving old code around, are there any changes or is this identical to the moved version? `has_cursor` is new, are there any other changes?

libtiny_tui/src/msg_area/line.rs
```diff
@@ -200,65 +187,41 @@ impl Line {
...
-
-        let mut char_idx: i32 = 0;
+        let mut line_num = 0;
+        let mut char_idx = 0;
+        let mut split_indices_iter = self.line_data.get_splits().iter().copied().peekable();
 
-        let last_seg: [&Seg; 1] = [&self.current_seg];
-        for seg in self.segs.iter().chain(last_seg.iter().copied()) {
+        let last_seg: [&StyledString; 1] = [&self.current_seg];
+        for seg in self.string_segments.iter().chain(last_seg.iter().copied()) {
```
osa1, line 197: No need to copy `last_seg` here

```suggestion
        for seg in self.string_segments.iter().chain(::std::iter::once(&self.current_seg)) {
```

Tbh I don't understand why `chain` here needs an iterator that owns the data, but it's not too important.

libtiny_tui/src/msg_area/line.rs
```diff
@@ -361,7 +324,7 @@ mod tests {
     fn height_test_5() {
         let mut line = Line::new();
         line.add_text("abcde");
-        for i in 0..6 {
+        for i in 6..0 {
```
osa1, line 327: Is there a reason for this change?

libtiny_tui/src/msg_area/line.rs
```diff
@@ -307,7 +270,7 @@ mod tests {
     fn height_test_1() {
         let mut line = Line::new();
         line.add_text("a b c d e");
-        assert_eq!(line.rendered_height(1), 5);
+        assert_eq!(line.rendered_height(1), 9);
```
osa1, line 273: :+1: I thought maybe we should preserve the old behavior here, but then if I see something like

```
aaaaa|
aaaaa|
```

There's no way to know whether that's `aaaaaaaaaa` split into two lines or `aaaaa aaaaa`. If we preserve the space here then we render it as

```
aaaaa|
 aaaa|
a    |
```

Now it's clear that the original string is `aaaaa aaaaa`.

libtiny_tui/src/line_split.rs
```diff
@@ -0,0 +1,152 @@
...
+        self.last_whitespace_idx = None;
+        self.prev_char_is_whitespace = false;
+        self.current_line_length = 0;
+    }
+
+    pub fn get_line_count(&self) -> Option<usize> {
+        self.line_count.map(|c| c as usize)
+    }
+
+    pub fn get_splits(&self) -> &Vec<i32> {
```
osa1, line 76:
```suggestion
    pub fn get_splits(&self) -> &[i32] { 
```

libtiny_tui/src/line_split.rs
trevarj: This is all identical except the addition of not having a cursor to account for.

libtiny_tui/src/msg_area/line.rs
trevarj: Thanks for the insight. Will remove.

libtiny_tui/src/msg_area/line.rs
trevarj: The original test doesn't meet the requirement. Now that I look closer at it, this one doesn't either because it's caching the first value. I think it's ok to remove it.

osa1: Which requirement do you mean?

trevarj: The long, spaceless lines no longer get cut off. 
Before it took "abcde" and rendered it cut off.
Ex. 
When i/width is 1, the string renders as "a" and the line count is 1.
width = 2 -> "ab", still 1 line
width = 3 -> "abc"...

My tests - `test_calculate_height()`- in `input_line.rs` cover cases like this.

trevarj:
Rebased with master.

PR #220, opened 2020-06-25
Added optional config to limit the number of messages in each tab.

### LINKED ISSUE #219: Is there a max buffer size?
Nice client!

I'm using tiny on a multi-user system in tmux, and i want to make sure I don't use an unbounded amount of memory if I just leave it open all day.

trevarj:
Closes #219.

### DIFF
--- a/libtiny_tui/src/config.rs
+++ b/libtiny_tui/src/config.rs
@@ -13,6 +13,8 @@ pub use termbox_simple::*;
 pub(crate) struct Config {
     #[serde(default)]
     pub(crate) colors: Colors,
+    #[serde(default = "usize::max_value")]
+    pub(crate) scrollback: usize,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
--- a/libtiny_tui/src/msg_area/mod.rs
+++ b/libtiny_tui/src/msg_area/mod.rs
@@ -1,14 +1,15 @@
 pub(crate) mod line;
 
+use std::collections::VecDeque;
 use std::{cmp::max, mem, str};
-
 use termbox_simple::Termbox;
 
 pub(crate) use self::line::{Line, SegStyle};
 use crate::config::Colors;
 
 pub(crate) struct MsgArea {
-    lines: Vec<Line>,
+    lines: VecDeque<Line>,
+    scrollback: usize,
 
     // Rendering related
     width: i32,
@@ -27,9 +28,10 @@ pub(crate) struct MsgArea {
 }
 
 impl MsgArea {
-    pub(crate) fn new(width: i32, height: i32) -> MsgArea {
+    pub(crate) fn new(width: i32, height: i32, scrollback: usize) -> MsgArea {
         MsgArea {
-            lines: Vec::new(),
+            lines: VecDeque::with_capacity(512.min(scrollback)),
+            scrollback,
             width,
             height,
             scroll: 0,
@@ -157,8 +159,13 @@ impl MsgArea {
 
     pub(crate) fn flush_line(&mut self) -> usize {
         let line_height = self.line_buf.rendered_height(self.width);
+        // Check if we're about to overflow
+        if self.lines.len() == self.scrollback {
+            // Remove oldest line
+            self.lines.pop_front();
+        }
         self.lines
-            .push(mem::replace(&mut self.line_buf, Line::new()));
+            .push_back(mem::replace(&mut self.line_buf, Line::new()));
         if self.scroll != 0 {
             self.scroll += line_height;
         }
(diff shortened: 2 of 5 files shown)


tiny/config.yml
```diff
@@ -39,6 +39,9 @@ defaults:
 # Where to put log files
 log_dir: '{}'
 
+# (Optional) Limits the maximum number of messages stored per each UI tab
+# max_buffer_per_tab: 512
```
trevarj, line 43: Could leave this uncommented and just blank. I believe it makes it None

libtiny_tui/src/tui.rs
```diff
@@ -50,6 +50,9 @@ pub(crate) struct TUI {
     /// Color scheme
     colors: Colors,
 
+    /// Max size of message lines
```
dontlaugh, line 53: This might be more clear:
```suggestion
    /// Max number of message lines
```

tiny/config.yml
dontlaugh: In the example config, we should note that the default is unlimited.

dontlaugh:
A couple of nits from me. I'll let people more familiar with the code comment. Thanks for jumping on this! My server admin thanks you! :+1:

libtiny_tui/src/tui.rs
trevarj: Thanks for the feedback. I wish I knew I could click "Commit suggestion". Never used that before!

dontlaugh: It's neat, right?!

tiny/config.yml
osa1: Let's call this "scrollback" or "history". The word "scrollback" is used in konsole, hexchat, konversation and probably many other programs, and in konversation and hexchat they don't even explain what it is about because it's so obvious. "history" is used by e.g. tmux.

Personally I prefer "scrollback" as that's the word I use to refer to this concept all the time, so it's familiar to me.

Also a nit: s/per each/for each

libtiny_tui/src/config.rs
```diff
@@ -13,6 +13,8 @@ pub use termbox_simple::*;
 pub(crate) struct Config {
     #[serde(default)]
     pub(crate) colors: Colors,
+    #[serde(default)]
+    pub(crate) max_buffer_per_tab: Option<usize>,
```
osa1, line 17: My renaming suggestion for this field is above.

I think we could use `usize` here and set it `usize::MAX` by default. That also simplifies the comparison below.

After renaming this let's also rename other parameter and field names so that they're all the same.

libtiny_tui/src/msg_area/mod.rs
```diff
@@ -159,6 +161,12 @@ impl MsgArea {
         let line_height = self.line_buf.rendered_height(self.width);
         self.lines
             .push(mem::replace(&mut self.line_buf, Line::new()));
+        // Remove first element because we went over the maximum lines
+        if let Some(max_lines) = self.max_lines {
```
osa1, line 165: If you simplify the `Option<usize>` type this will be gone and the line below will be `if self.lines.len() > self.max_lines { .. }`.

libtiny_tui/src/msg_area/mod.rs
```diff
@@ -159,6 +161,12 @@ impl MsgArea {
         let line_height = self.line_buf.rendered_height(self.width);
         self.lines
             .push(mem::replace(&mut self.line_buf, Line::new()));
```
osa1, line 163: Such a minor thing, but if you check the length, then pop when necessary and then push, I think you can save some allocation here. No idea how often it'll happen in practice, but it just amounts to replacing two statements so I'd do it.

In other words, instead of

- Push new item
- Check if it overflowed, pop one item if it did

Do

- Check if push will overflow and pop an item if it will
- Push item

libtiny_tui/src/msg_area/mod.rs
```diff
@@ -159,6 +161,12 @@ impl MsgArea {
...
+        // Remove first element because we went over the maximum lines
+        if let Some(max_lines) = self.max_lines {
+            if self.lines.len() > max_lines {
+                self.lines.remove(0);
```
osa1, line 167: Why not use `VecDeque` here? This potentially (depending of buffer size) shifts thousands of elements.

libtiny_tui/src/msg_area/mod.rs
```diff
@@ -27,9 +28,10 @@ pub(crate) struct MsgArea {
 }
 
 impl MsgArea {
-    pub(crate) fn new(width: i32, height: i32) -> MsgArea {
+    pub(crate) fn new(width: i32, height: i32, scrollback: usize) -> MsgArea {
         MsgArea {
-            lines: Vec::new(),
+            lines: VecDeque::with_capacity(512),
```
trevarj, line 33: Thoughts on making this `512.min(scrollback)` ?

osa1: I think we should do this, yes.

osa1:
Thanks!

PR #221, opened 2020-06-26
Remove editor support

osa1:
See comments in code and the changelog entry for details.

It's hard to support this feature, and I'm not sure that it's used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,35 +1,47 @@
 # Unreleased
 
+## New features
+
+- It's now possible to build tiny with [rustls] instead of [native-tls]. See
+  README for instructions. (#172)
+- A new optional server field 'alias' added to the configuration file for
+  specifying aliases for servers, to be shown in the tab line. This is useful
+  when a server address is long, or just an IP address, or you want to show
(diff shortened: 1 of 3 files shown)


Kabouik:
Could you please give any pointer on what should be my git workflow if I want to keep this feature when updating `tiny` in the future? I missed that PR before but I definitely use `$EDITOR` a lot when using `tiny`.

trevarj:
So this auto-sends all lines except last? It makes me nervous if I have some long text in my clipboard and I accidentally right click/middle click the terminal window and send a bunch of lines.

@Kabouik I think you could create a branch from the commit immediately before this change's commit (once it's in master), then you can `rebase -i` to master and `drop` the commit that removes $EDITOR when you want to get new changes.

osa1:
OK, maybe removing this is not a good idea.

Let's try to find the least hacky way to fix this feature. To rephrase: the problem is we need to stop handling stdin when we run `$EDITOR` so that the editor can read it. For this we'll either need a way to "unregister" `term_input` from the event loop, or somehow make `term_input` not read stdin (e.g. it'll still be registered, but until we "activate" it again it won't ever be ready for reading).

PR #222, opened 2020-06-28
Fix $EDITOR support

### LINKED ISSUE #185: Panic in pinger.rs

osa1:
Previously running $EDITOR would block tokio event loop, which causes
all kinds of panics, and timeouts in connections, see #185 as an
example.

We now run $EDITOR in a new thread, and stop reading stdin in the input
handler task during that. Also fixes termbox's suspend method, which
previously keps using tty after suspend.

See also module documentation of `editor`.

Fixes #185

### DIFF
--- a/libtiny_tui/src/editor.rs
+++ b/libtiny_tui/src/editor.rs
@@ -1,35 +1,53 @@
+//! This module implements some part of running $EDITOR to edit input. The other part is
+//! implemented at the top-level of the current crate (in `input_handler` task).
+//!
+//! Implementation is a little bit hacky: we need to give control of tty, stdout, and stdin to
+//! $EDITOR while still running the tokio event loop, to handle connection events. Here's how we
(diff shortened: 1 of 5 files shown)


osa1:
I tried this by running `$EDITOR` for about 5 minutes while being connected to 4 servers, and nothing broke! Merging.

PR #223, opened 2020-06-28
InputArea/LineDataCache, fast remove character from end of input

### LINKED ISSUE #208: Input area improvements/tasks

trevarj:
This is the final requirement to close #208.

> Deleting from the end shouldn't require generating the state from scratch.

I added a function to `LineDataCache` called `reverse_by_one`, which rewinds the state by one "step" in the calculation of height (`calculate_height`). 

I had to add a field to `LineDataCache`, `line_lengths`, which stores the length of each line. These can vary in size due word wrapping.

The reversal is pretty procedural with a "fast-track" when there is only one line to begin with. There might be a way to simplify the conditional logic, but I haven't explored that due to added complexity (and because I don't know if there is a way to make it simpler :smile:).

### DIFF
--- a/libtiny_tui/src/line_split.rs
+++ b/libtiny_tui/src/line_split.rs
@@ -13,12 +13,14 @@ pub struct LineDataCache {
     /// Current nickname length. Used in determining if we need to invalidate due to resize.
     nick_length: usize,
     /// The index into InputLine::buffer of the last whitespace that we saw in calculate_height()
-    last_whitespace_idx: Option<i32>,
+    last_whitespace_idx: Vec<Option<i32>>,
     /// True if the last character was a whitespace character
     prev_char_is_whitespace: bool,
     /// The length of the current line that is being added to.
     /// Used to determine when to wrap to the next line in calculate_height()
     current_line_length: i32,
+    /// A stack of each line length
+    line_lengths: Vec<i32>,
     /// If the line of text has a cursor
     has_cursor: bool,
 }
@@ -31,9 +33,10 @@ impl LineDataCache {
             width: 0,
             line_width: 0,
             nick_length: 0,
-            last_whitespace_idx: None,
+            last_whitespace_idx: Vec::new(),
             prev_char_is_whitespace: false,
             current_line_length: 0,
+            line_lengths: Vec::new(),
             has_cursor,
         }
     }
@@ -64,9 +67,10 @@ impl LineDataCache {
         self.width = width;
         self.nick_length = nick_length;
         self.line_width = width - nick_length as i32;
-        self.last_whitespace_idx = None;
+        self.last_whitespace_idx = Vec::new();
         self.prev_char_is_whitespace = false;
         self.current_line_length = 0;
+        self.line_lengths = Vec::new();
     }
 
     pub fn get_line_count(&self) -> Option<usize> {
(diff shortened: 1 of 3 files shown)


libtiny_tui/src/tests/mod.rs
```diff
@@ -412,7 +412,6 @@ fn test_text_field_wrap() {
         tui.handle_input_event(event, &mut None);
     }
     // InputLine cache gets invalidated after backspace, need to redraw to calculate.
-    tui.draw();
```
trevarj, line 415: No longer needed because backspacing at the end will reverse state properly.

libtiny_tui/src/line_split.rs
```diff
@@ -149,4 +160,121 @@ impl LineDataCache {
...
+        let mut temp_count = 1;
+        if let Some(line_count) = self.line_count {
+            temp_count = line_count;
+            // If we made space for the cursor, subtract it.
+            if self.has_cursor && self.current_line_length == self.line_width {
+                temp_count -= 1;
+            }
+        }
+        // Helper to go backwards and find the last whitespace index
+        fn find_last_whitespace_idx(buffer: &[char]) -> Option<i32> {
```
trevarj, line 177: Originally, I converted `last_whitespace_idx` to a `Vec<Option<i32>>` then ran `bench.rs` and got really bad results. Changing to this solved that.

osa1:
Thanks @trevarj. I briefly looked at this PR now. My only concern is it seems quite complicated so I'm not sure if it's worth adding. I wonder if it's possible to make this simpler somehow. I'm not confident that this much non-trivial code won't come with its own bugs, especially because we don't have good test coverage in this code.

Btw, I think this is a feature that could be tested using QuickCheck-style property testing. Generate a random initial string and a list of insert and remove events (make sure the insertions and removals are in range), then for each event (1) update the string manually using the String API (2) update the input buffer using the fast paths for insertion and removal (3) create a new input_line from scratch using the string in (1). Now if the code is correct contents of (2) and (3) should be the string (1), and (2) and (3) should have the same internal state.

This could be used to test the current implementation (without this patch) as well.

Just an idea, not asking you to do it.

trevarj:
@osa1 Thanks for taking the time to look. I completely agree about the unnecessary complexity. I just wanted to do it regardless, to simply see if I could figure it out, but I'm ok with not merging this stuff. 

We can close this PR and perhaps make an issue for adding your test case idea for this area.

PR #224, opened 2020-06-29
CTCP VERSION message support

### LINKED ISSUE #145: Handle CTCP VERSION messages
I thought the "version" messages sent by `freenode-connect` are some non-standard thing, but apparently it's a widely supported CTCP message: https://tools.ietf.org/id/draft-oakley-irc-ctcp-01.html#version

We should handle these messages properly according to the RFC.

trevarj:
- Added `ClientInfo` to `Client` which stores the client's version.
- Added a  `/version <nick>` command that sends a request for that nick's version
- Handle the response to `/version`
- Now we respond to other user's asking us for which version we're on
___
Unrelated:
- Refactored some of the command usages
- Added `/help`

Closes #145

### DIFF
--- a/libtiny_client/examples/echo.rs
+++ b/libtiny_client/examples/echo.rs
@@ -1,6 +1,6 @@
 //! An echo bot that just repeats stuff sent to it (either in a channel or as PRIVMSG).
 
-use libtiny_client::{Client, Event, ServerInfo};
+use libtiny_client::{Client, ClientInfo, Event, ServerInfo};
 use libtiny_wire::{Cmd, Msg, MsgTarget, Pfx};
 
 use futures::stream::StreamExt;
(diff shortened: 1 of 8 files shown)


trevarj:
@osa1 what do you think about having this for v1.0?

osa1:
> @osa1 what do you think about having this for v1.0?

I don't understand why would anyone want to expose any of their details to strangers on IRC, it seems like a bad idea to me. Is this necessary? Or useful?

trevarj:
I can sort of agree with that.  Could it potentially be an attack vector? 😅 

For some reason it is a "MUST" implemented CTCP message... 
https://tools.ietf.org/id/draft-oakley-irc-ctcp-01.html#version
> VERSION is universally implemented. Clients MUST implement this CTCP message.

osa1:
> For some reason it is a "MUST" implemented CTCP message...

That's funny... to me it seems useless at best. Unless servers start to ban users that don't respond to `VERSION` I see no reason add more code for this.

(If servers did this I think we could just respond to servers and not to other users)

PR #226, opened 2020-06-30
Fixes panic when pasting a long string.

### LINKED ISSUE #225: Panic when pasting long string
I just got this after I pasted a long, single-line string using `C-x`:

```
thread 'main' panicked at 'index out of bounds: the len is 5439 but the index is 5439', termbox/src/lib.rs:350:18
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:217
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:520
  11: rust_begin_unwind
             at src/libstd/panicking.rs:431
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  13: core::panicking::panic_bounds_check
             at src/libcore/panicking.rs:62
  14: libtiny_tui::input_area::input_line::draw_line_wrapped
  15: libtiny_tui::input_area::input_line::draw_line
  16: libtiny_tui::tui::TUI::draw
  17: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  18: tokio::runtime::task::raw::poll
  19: tokio::task::local::LocalSet::tick
  20: std::thread::local::LocalKey<T>::with
  21: tiny::main
  22: std::rt::lang_start_internal::{{closure}}::{{closure}}
             at src/libstd/rt.rs:52
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I think this happens when the pasted string takes multiple lines on the screen (using wrapping). cc @trevarj

trevarj:
When pasting a multiline string, specifically when coming back from the editor, we need to call recalculate so the input line is drawn at the correct height, otherwise it will go out of bounds.

Closes #225

### DIFF
--- a/libtiny_tui/src/input_area/mod.rs
+++ b/libtiny_tui/src/input_area/mod.rs
@@ -460,6 +460,7 @@ impl InputArea {
     pub(crate) fn set(&mut self, str: &str) {
         self.mode = Mode::Edit;
         self.buffer = InputLine::from_buffer(str.chars().collect());
+        self.height = None;
         self.move_cursor_to_end();
     }
 
@@ -803,5 +804,14 @@ mod tests {
         assert_eq!(text_field.cursor, 4);
         text_field.keypressed(Key::CtrlArrow(Arrow::Right));
         assert_eq!(text_field.cursor, 5);
+    }
+
+    #[test]
+    fn test_set_buffer() {
+        let mut input_area = InputArea::new(40, 50);
+        // a string that will be more than one line - 41 characters
+        let multiline_string_no_spaces = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+        input_area.set(&multiline_string_no_spaces);
+        assert_eq!(input_area.get_height(input_area.width), 2);
     }
 }


trevarj:
This was an oversight by me when writing the line wrapping code.

When set() gets called when coming back from the editor, InputArea already has a height set, so it won't try to recalculate. 

Another option would be to invalidate the height (`self.height = None`) instead of recalculating in set(). @osa1 which do you prefer?

osa1:
Great catch @trevarj, thanks!

> Another option would be to invalidate the height (self.height = None) instead of recalculating in set(). @osa1 which do you prefer?

I think I prefer `self.height = None` and delaying calculating height as much as possible. Reason: if there are other code that also invalidates the height after editor returns, but before the next `draw` call (e.g. maybe we decide to fall back to scrolling) we avoid redundantly calculating height here only to invalidate it again before the next `draw` call.

trevarj:
Changed to `self.height = None`, and squashed my commits.

osa1:
Thanks!

PR #228, opened 2020-07-01
Fixed join/part line modification panic

### LINKED ISSUE #227: Join/Part messages overflow
Saw this randomly when a bunch of people all joined/parted at the same time. It somehow corrected itself eventually, but still worth some investigation.

image

trevarj:
When multiple users join/part from a channel at the same timestamp, the line gets modified. This wasn't invalidating the height calulation so it was causing an overflow if many people join/part or if your screen is small and it needs to wrap that line after modification.

Ugh, basically what I did for a fix is just add a way to force invalidating the height of the line when it's modified. Didn't think about this when I added `LineDataCache` for height calculation.

Closes #227

### DIFF
--- a/libtiny_tui/src/msg_area/line.rs
+++ b/libtiny_tui/src/msg_area/line.rs
@@ -149,8 +149,12 @@ impl Line {
         self.len_chars += 1;
     }
 
-    /// How many lines does this take when rendered? O(n) where n = number of
-    /// split positions in the line (i.e. whitespaces).
+    pub(crate) fn force_recalculation(&mut self) {
+        self.line_data.set_dirty()
+    }
+
+    /// Calculates the number of lines that this line will be.
+    /// The calculation is only done if the line_data is dirty or the window is resized.
     pub(crate) fn rendered_height(&mut self, width: i32) -> i32 {
         if self.line_data.is_dirty() || self.line_data.needs_resize(width, 0) {
             self.line_data.reset(width, 0);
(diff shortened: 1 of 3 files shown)


trevarj:
Closes #227

libtiny_tui/src/msg_area/line.rs
```diff
@@ -149,8 +149,12 @@ impl Line {
         self.len_chars += 1;
     }
 
-    /// How many lines does this take when rendered? O(n) where n = number of
-    /// split positions in the line (i.e. whitespaces).
+    pub(crate) fn force_recalculation(&mut self) {
```
trevarj, line 152: Not sure if this actually needs to be `pub(crate)`. `pub` might be the right choice, but I'm not sure if matters.

libtiny_tui/src/msg_area/line.rs
```diff
@@ -149,8 +149,12 @@ impl Line {
         self.len_chars += 1;
     }
 
-    /// How many lines does this take when rendered? O(n) where n = number of
-    /// split positions in the line (i.e. whitespaces).
```
osa1, line 153: Why was this comment removed? Is this not valid anymore?

osa1:
LGTM, thanks @trevarj ! Added one inline question.

libtiny_tui/src/msg_area/line.rs
trevarj: Yeah, it's O(n) where n is the size of the buffer

PR #233, opened 2020-08-12
Added help command (/help).

trevarj:
Since there are two areas for sending commands, in `libtiny_tui` and in the tiny client, I had to add some changes in two areas.

I aligned the /help message because I personally found it difficult to read. I can also imagine a user typing /help and wanting to have the commands printed out nicely so a command can be easily found.

image

### DIFF
--- a/libtiny_tui/src/tui.rs
+++ b/libtiny_tui/src/tui.rs
@@ -38,6 +38,42 @@ pub(crate) enum TUIRet {
 const LEFT_ARROW: char = '<';
 const RIGHT_ARROW: char = '>';
 
+struct CmdUsage {
+    name: &'static str,
+    description: &'static str,
+    usage: &'static str,
+}
+
+impl CmdUsage {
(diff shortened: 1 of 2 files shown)


libtiny_tui/src/tui.rs
```diff
@@ -38,6 +38,28 @@ pub(crate) enum TUIRet {
 const LEFT_ARROW: char = '<';
 const RIGHT_ARROW: char = '>';
 
+/// Command - (name, description, usage)
+struct CmdUsage(&'static str, &'static str, &'static str);
```
trevarj, line 42: This unit struct is probably stupid, but at the time I thought it was nice to create all the following constants...

osa1: I think this is fine, but perhaps we can give names to fields so that the type will become self-documenting. Something like

```rust
struct CmdUsage {
    name: &'static str,
    description: &'static str,
    usage: &'static str,
}
```

Now we can remove the documentation as it's clear from the type.

trevarj: Done!

PR #237, opened 2020-08-17
Restore away msg, /names sort

### LINKED ISSUE #234: Set away status on reconnect
See the TODO in the source code: https://github.com/osa1/tiny/blob/efcb30310e0113c51e27cb21bbbd8e6783f064a7/libtiny_client/src/state.rs#L386

Basically we don't set away status on reconnect, so, if a user is away but loses connection, on reconnect they won't be away anymore.

Fix is easy: just send an away message after the join messages in the linked code.

### LINKED ISSUE #235: Lexicographically sort /names list
This feature request came from a user in tiny IRC channel recently. I think it makes sense, but we will probably need to add a dependency for this (probably [lexical_sort](https://docs.rs/lexical-sort/0.3.0/lexical_sort/)).

### LINKED ISSUE #236: Wait for NickServ response before joining channels when nickserv_ident is used
We currently send NickServ identify message after RPL_WELCOME and join message after RPL_ENDOFMOTD. RPL_ENDOFMOTD is sent after RPL_WELCOME, but sometimes NickServ takes longer and authentication isn't successful until after RPL_ENDOFMOTD. When that happens we end up sending JOIN message before authenticating, which fails for channels with `+R` (login-to-join).

When nickserv_ident is used we should wait to hear from NickServ before sending JOIN.

Implementation is simple: just check for PRIVMSGs from NickServ in `update` in `libtiny_client/src/state.rs` and send join messages there. We'll need a `bool` field for whether we've sent JOIN already as we don't want to send JOIN every time we get a PRIVMSG from NickServ. The state should be `false` initially and reset on disconnect, and should be `true` after sending the JOIN message.

trevarj:
Restore away status if reconnected. Closes #234.

~~If a channel is restricted and we can't autojoin until identified with NickServ, we retry joining on a 477 response from the server. Closes #236~~

Sorted nickname list on /names command lexigraphically. Requires new
crate - lexical-sort. Closes #235.

Cleaned up some TODOs by adding logs in state.rs.

### DIFF
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -383,12 +383,17 @@ impl StateInner {
             }
 
             //
-            // RPL_ENDOFMOTD, join channels, set away status (TODO)
+            // RPL_ENDOFMOTD, join channels, set away status
             //
             Reply { num: 376, .. } => {
                 let chans: Vec<&str> = self.chans.iter().map(|(s, _)| s.as_str()).collect();
                 if !chans.is_empty() {
                     snd_irc_msg.try_send(wire::join(&chans)).unwrap();
+                }
+                if self.away_status.is_some() {
+                    snd_irc_msg
+                        .try_send(wire::away(self.away_status.as_deref()))
+                        .unwrap();
                 }
             }
 
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -70,9 +70,8 @@ struct StateInner {
     /// Nicks to try, in this order.
     nicks: Vec<String>,
 
-    /// NickServ passowrd.
+    /// NickServ password
     nickserv_ident: Option<String>,
-
     /// An index to `nicks`. When out of range we add `current_nick_idx - nicks.length()`
     /// underscores to the last nick in `nicks`
     current_nick_idx: usize,
@@ -233,6 +232,17 @@ impl StateInner {
                 }
             }
 
+            // Reply 477 is sent when user needs to be identified with NickServ to join a channel
+            // Only try re-join if nickserv_ident is configured
+            Reply { num: 477, params } => {
+                if self.nickserv_ident.is_some() {
+                    if let Some(chan) = params.get(1) {
+                        debug!("Attempting to re-join channel {}", chan);
+                        snd_irc_msg.try_send(wire::join(&[chan])).unwrap();
+                    }
+                }
+            }
+
             Reply { num: 396, params } => {
                 // :hobana.freenode.net 396 osa1 haskell/developer/osa1
                 // :is now your hidden host (set by services.)
@@ -255,7 +265,7 @@ impl StateInner {
                 let param = &params[1];
                 match param.find('=') {
                     None => {
-                        // TODO: Log this
+                        error!("Could not parse 302 RPL_USERHOST to set usermask.");
                     }
                     Some(mut i) => {
                         if param.as_bytes().get(i + 1) == Some(&b'+')
@@ -323,7 +333,7 @@ impl StateInner {
 
                 match parse_servername(params) {
                     None => {
-                        // TODO: Log
+                        error!("Could not parse server name in 002 RPL_YOURHOST message.");
                     }
                     Some(servername) => {
                         self.servername = Some(servername);
@@ -426,7 +436,7 @@ impl StateInner {
                 let nick = match pfx {
                     Some(Pfx::User { nick, .. }) => nick,
                     _ => {
-                        // TODO: WAT?
+                        error!("Could not extract nick from QUIT message.");
                         return;
                     }
                 };
@@ -496,7 +506,10 @@ impl StateInner {
 
     fn get_chan_nicks(&self, chan: &str) -> Vec<String> {
         match utils::find_idx(&self.chans, |(s, _)| s == chan) {
-            None => vec![], // TODO: Log this, this is probably a bug
+            None => {
+                warn!("Could not find channel index in get_chan_nicks.");
+                vec![]
+            }
             Some(chan_idx) => self.chans[chan_idx].1.iter().cloned().collect(),
         }
     }
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -232,17 +232,6 @@ impl StateInner {
                 }
             }
 
-            // Reply 477 is sent when user needs to be identified with NickServ to join a channel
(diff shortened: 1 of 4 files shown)


tiny/src/cmd.rs
```diff
@@ -422,7 +422,9 @@ fn names(args: CmdArgs) {
     };
 
     if let MsgSource::Chan { ref serv, ref chan } = src {
-        let nicks_vec = client.get_chan_nicks(chan);
+        use lexical_sort::{lexical_only_alnum_cmp, StringSort};
```
trevarj, line 425: Scoped 'use...' to limit the scope of the StringSort trait? Not sure if this is an anti-pattern or accepted as "efficient" or something.

libtiny_client/src/state.rs
```diff
@@ -255,7 +265,7 @@ impl StateInner {
                 let param = &params[1];
                 match param.find('=') {
                     None => {
-                        // TODO: Log this
+                        error!("Could not parse 302 RPL_USERHOST to set usermask.");
```
osa1, line 268: Following https://stackoverflow.com/a/2031209/691032 I think this should be `warn` because it doesn't break anything, it's just an interesting case.

libtiny_client/src/state.rs
```diff
@@ -323,7 +333,7 @@ impl StateInner {
 
                 match parse_servername(params) {
                     None => {
-                        // TODO: Log
+                        error!("Could not parse server name in 002 RPL_YOURHOST message.");
```
osa1, line 336: This case is more interesting, and I think I found a potential bug while looking at uses of `self.servername`. Currently we use the server name when sending pings, and we don't send pings if we're unable to parse the server name from RPL_YOURHOST, but sending pings is pretty important as that's how we check connectivity. I think if we're unable to parse server name we can use the server address as the ping target. Or maybe just send a PING without an argument and wait for an error message from the server (which should reset the ping timer). For example, this works fine to check connectivity:

```
>>> PING <-- sent message
13:33 cherryh.freenode.net: osa1 PING Not enough parameters
```

When we receive this error message the pinger timer will be reset.

Only concern is if we can't parse a server name and send PING messages without a target then the server tab will have lots of "PING not enough parameters" lines. Maybe that's fine though as this should be a rare case.

For this PR I think it's fine to add an error-level log here.

libtiny_client/src/state.rs
```diff
@@ -383,13 +393,18 @@ impl StateInner {
...
             Reply { num: 376, .. } => {
                 let chans: Vec<&str> = self.chans.iter().map(|(s, _)| s.as_str()).collect();
                 if !chans.is_empty() {
                     snd_irc_msg.try_send(wire::join(&chans)).unwrap();
                 }
+                if self.away_status.is_some() {
+                    snd_irc_msg
+                        .try_send(wire::away(self.away_status.as_deref()))
+                        .unwrap();
+                }
```
osa1, line 407: Thanks!

libtiny_client/src/state.rs
```diff
@@ -421,7 +436,7 @@ impl StateInner {
                 let nick = match pfx {
                     Some(Pfx::User { nick, .. }) => nick,
                     _ => {
-                        // TODO: WAT?
+                        error!("Could not extract nick from QUIT message.");
```
osa1, line 439: Let's make this warn as well.

libtiny_client/src/state.rs
```diff
@@ -491,7 +506,10 @@ impl StateInner {
 
     fn get_chan_nicks(&self, chan: &str) -> Vec<String> {
         match utils::find_idx(&self.chans, |(s, _)| s == chan) {
-            None => vec![], // TODO: Log this, this is probably a bug
+            None => {
+                warn!("Could not find channel index in get_chan_nicks.");
```
osa1, line 510: This should be `error` as it indicates a bug in tiny.

tiny/src/cmd.rs
osa1: I think this is fine.

Did you check how much the release binary size changes with this library?

osa1:
Here's a screenshot of me trying to join a channel that requires registration but my nick is not registered and `nickserv_ident` is specified:

Screenshot_2020-08-18_05-21-15

This goes on forever. I guess we'll need to stop trying after a few attempts. It'd also be good to have some delay between attempts.

I wonder how sopel handles this?

@trevarj also if you split this into two PRs we can merge the rest of the PR.

trevarj:
Removed changes for #236 to merge other changes.

osa1:
For the record, with the new dependency tiny binary sizes goes from 6372912 bytes to 6865872 bytes. That's +0.5M, +7.7%.

osa1:
Here's an idea: IRC is an ASCII-based protocol, so nicks are supposed to be ASCII. I just tried using a non-ASCII nick and it's rejected:

```
helix.oftc.net: osa1 ömer Erroneous Nickname
```

I think this is the correct behavior.

If we assume ASCII strings then we don't need new dependencies, we could simply sort using ASCII codes. Only tricky part is lower and upper case letters, which should be easy enough to handle without requiring 0.5M of dependencies.

trevarj:
> Here's an idea: IRC is an ASCII-based protocol, so nicks are supposed to be ASCII. I just tried using a non-ASCII nick and it's rejected:
> 
> ```
> helix.oftc.net: osa1 ömer Erroneous Nickname
> ```
> 
> I think this is the correct behavior.
> 
> If we assume ASCII strings then we don't need new dependencies, we could simply sort using ASCII codes. Only tricky part is lower and upper case letters, which should be easy enough to handle without requiring 0.5M of dependencies.

Ok, great. I reverted and now I'm using regular sort on Strings.

libtiny_client/src/state.rs
```diff
@@ -491,8 +495,21 @@ impl StateInner {
 
     fn get_chan_nicks(&self, chan: &str) -> Vec<String> {
         match utils::find_idx(&self.chans, |(s, _)| s == chan) {
-            None => vec![], // TODO: Log this, this is probably a bug
-            Some(chan_idx) => self.chans[chan_idx].1.iter().cloned().collect(),
+            None => {
+                error!("Could not find channel index in get_chan_nicks.");
+                vec![]
+            }
+            Some(chan_idx) => {
```
trevarj, line 502: Sorting here now to leave the logic out of the cmd.rs module.
Using unstable because it should be faster and we don't need to worry about equal elements (when it becomes unstable)

osa1:
Thanks!

PR #240, opened 2020-08-19
Changes to state to handle auto-joining channels that have +R.

### LINKED ISSUE #236: Wait for NickServ response before joining channels when nickserv_ident is used

trevarj:
If the server sends back a 477 reply, meaning that we can't join the
channel without being registered and identified with NickServ, then we
check to see if nickserv_ident is configured. If it is configured then
we wait for 10 seconds and retry joining the channel. We do this 3 times
and then stop.

Also cleaned up channel tuple and made it into a struct so that I could add a retry count for each channel.
Closes #236

### DIFF
--- a/libtiny_client/src/lib.rs
+++ b/libtiny_client/src/lib.rs
@@ -100,6 +100,8 @@ pub enum Event {
     Msg(wire::Msg),
     /// A wire-protocol error
     WireError(String),
+    /// Channel join error message
+    ChannelJoinError { chan: String, msg: String },
 }
 
 impl From<StreamError> for Event {
@@ -220,7 +222,7 @@ impl Client {
 
     /// Leave a channel.
     pub fn part(&mut self, chan: &str) {
-        self.msg_chan.try_send(Cmd::Msg(wire::part(chan))).unwrap();
+        self.state.leave_channel(&mut self.msg_chan, chan)
     }
 
     /// Set away status. `None` means not away.
@@ -242,6 +244,8 @@ impl Client {
     /// outgoing messages) will be dropped.
     pub fn quit(&mut self, reason: Option<String>) {
         debug!("quit cmd received");
+        // Kill any retry join tasks to prevent quit delay
+        self.state.kill_join_tasks();
         self.msg_chan.try_send(Cmd::Quit(reason)).unwrap();
     }
 
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -8,11 +8,9 @@ use libtiny_wire::{Msg, Pfx};
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::rc::Rc;
-use tokio::{
-    select,
-    stream::StreamExt,
-    sync::mpsc::{Receiver, Sender},
-};
+
+use futures::{select, FutureExt, StreamExt};
+use tokio::sync::mpsc::{Receiver, Sender};
 
 #[derive(Clone)]
 pub struct State {
@@ -699,17 +697,21 @@ impl StateInner {
 async fn retry_channel_join(
     channel: String,
     mut snd_irc_msg: Sender<String>,
-    mut rcv_abort: Receiver<()>,
+    rcv_abort: Receiver<()>,
 ) {
     debug!("Attempting to re-join channel {}", channel);
+
     use tokio::time::{delay_for, Duration};
-    let mut delay = delay_for(Duration::from_secs(10));
+
+    let mut delay = delay_for(Duration::from_secs(10)).fuse();
+    let mut rcv_abort = rcv_abort.fuse();
+
     select! {
-        () = &mut delay => {
+        () = delay => {
             // Send join message
             snd_irc_msg.try_send(wire::join(&[&channel])).unwrap();
         },
-        Some(()) = rcv_abort.next() => {
+        _ = rcv_abort.next() => {
             // Channel tab was closed
         },
     };
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -295,7 +295,10 @@ impl StateInner {
                         // Initialize channel state
                         match utils::find_idx(&self.chans, |c| &c.name == chan) {
                             None => {
-                                self.chans.push(Chan::new(chan.to_owned()));
+                                let mut chan = Chan::new(chan.to_owned());
+                                // Since nick was found in the prefix, we are in the channel
+                                chan.join_state = JoinState::Joined;
+                                self.chans.push(chan);
                             }
                             Some(chan_idx) => {
                                 // This happens because we initialize channel states for channels
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -1,14 +1,18 @@
 #![allow(clippy::zero_prefixed_literal)]
 
 use crate::utils;
-use crate::{Event, ServerInfo};
+use crate::{Cmd, Event, ServerInfo};
 use libtiny_wire as wire;
 use libtiny_wire::{Msg, Pfx};
 
 use std::cell::RefCell;
 use std::collections::HashSet;
 use std::rc::Rc;
-use tokio::sync::mpsc::Sender;
+use tokio::{
+    select,
+    stream::StreamExt,
+    sync::mpsc::{Receiver, Sender},
+};
 
 #[derive(Clone)]
 pub struct State {
@@ -64,6 +68,14 @@ impl State {
     pub(crate) fn get_chan_nicks(&self, chan: &str) -> Vec<String> {
         self.inner.borrow().get_chan_nicks(chan)
     }
+
+    pub(crate) fn leave_channel(&self, msg_chan: &mut Sender<Cmd>, chan: &str) {
+        self.inner.borrow_mut().leave_channel(msg_chan, chan)
+    }
+
+    pub(crate) fn kill_join_tasks(&self) {
+        self.inner.borrow_mut().kill_join_tasks();
+    }
 }
 
 struct StateInner {
@@ -88,7 +100,7 @@ struct StateInner {
     ///
     /// TODO: I'm not sure if this is necessary. Why not just create channel tabs in the specified
     /// order, in TUI?
-    chans: Vec<(String, HashSet<String>)>,
+    chans: Vec<Chan>,
 
     /// Away reason if away mode is on. `None` otherwise. TODO: I don't think the message is used?
     away_status: Option<String>,
@@ -109,13 +121,89 @@ struct StateInner {
     server_info: ServerInfo,
 }
 
+#[derive(Debug)]
+struct Chan {
+    /// Name of the channel
+    name: String,
+    /// Set of nicknames in channel
+    nicks: HashSet<String>,
+    /// Channel joined state
+    join_state: JoinState,
+    /// Join attempts
+    join_attempts: u8,
+}
+
+/// State transitions:
+///    NotJoined -> Joining: When we get 477 for the channel
+///    NotJoined -> Joined: When we get a JOIN message for the channel on first attempt
+///    Joining -> Joined: When we get a JOIN message for the channel
+///    Joining -> NotJoined: Connection reset
+///    Joined -> NotJoined: Connection reset
+///    Joined -> Joining: Unexpected/Invalid state
+#[derive(Debug)]
+enum JoinState {
+    /// Initial state for Chan
+    NotJoined,
+    /// In the process of joining the channel
+    Joining {
+        /// Sender to kill the retry task if tab is closed
+        stop_task: Sender<()>,
+    },
+    /// Successfully joined the channel
+    Joined,
+}
+
+const MAX_JOIN_RETRIES: u8 = 3;
+
+impl Chan {
+    fn new(name: String) -> Chan {
+        Chan {
+            name,
+            nicks: HashSet::new(),
+            join_state: JoinState::NotJoined,
+            join_attempts: MAX_JOIN_RETRIES,
+        }
+    }
+
+    fn with_nicks(name: String, nicks: HashSet<String>) -> Chan {
+        Chan {
+            name,
+            nicks,
+            join_state: JoinState::NotJoined,
+            join_attempts: MAX_JOIN_RETRIES,
+        }
+    }
+
+    fn reset(&mut self) {
+        self.nicks.clear();
+        self.join_state = JoinState::NotJoined;
+        self.join_attempts = MAX_JOIN_RETRIES;
+    }
+
+    fn set_joining(&mut self, stop_task: Sender<()>) {
+        self.join_state = JoinState::Joining { stop_task }
+    }
+
+    /// Uses a retry.
+    /// Returns number of retries left or None.
+    fn retry_join(&mut self) -> Option<u8> {
+        match self.join_attempts {
+            0 => None,
+            _ => {
+                self.join_attempts -= 1;
+                Some(self.join_attempts)
+            }
+        }
+    }
+}
+
 impl StateInner {
     fn new(server_info: ServerInfo) -> StateInner {
         let current_nick = server_info.nicks[0].to_owned();
         let chans = server_info
             .auto_join
             .iter()
-            .map(|s| (s.clone(), HashSet::new()))
+            .map(|s| Chan::new(s.to_owned()))
             .collect();
         StateInner {
             nicks: server_info.nicks.clone(),
@@ -137,8 +225,8 @@ impl StateInner {
         self.current_nick_idx = 0;
         self.current_nick = self.nicks[0].clone();
         // Only reset the values here; the key set will be used to join channels
-        for (_, chan) in &mut self.chans {
-            chan.clear();
+        for chan in &mut self.chans {
+            chan.reset();
         }
         self.servername = None;
         self.usermask = None;
@@ -207,21 +295,23 @@ impl StateInner {
                         self.usermask = Some(usermask);
 
                         // Initialize channel state
-                        match utils::find_idx(&self.chans, |(s, _)| s == chan) {
+                        match utils::find_idx(&self.chans, |c| &c.name == chan) {
                             None => {
-                                self.chans.push((chan.to_owned(), HashSet::new()));
+                                self.chans.push(Chan::new(chan.to_owned()));
                             }
                             Some(chan_idx) => {
                                 // This happens because we initialize channel states for channels
                                 // that we will join on connection when the client is first created
-                                self.chans[chan_idx].1.clear();
+                                let chan = &mut self.chans[chan_idx];
+                                chan.join_state = JoinState::Joined;
+                                chan.nicks.clear();
                             }
                         }
                     } else {
-                        match utils::find_idx(&self.chans, |(s, _)| s == chan) {
+                        match utils::find_idx(&self.chans, |c| &c.name == chan) {
                             Some(chan_idx) => {
                                 self.chans[chan_idx]
-                                    .1
+                                    .nicks
                                     .insert(wire::drop_nick_prefix(nick).to_owned());
                             }
                             None => {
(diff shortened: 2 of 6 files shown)


libtiny_client/src/state.rs
```diff
@@ -88,7 +88,7 @@ struct StateInner {
     ///
     /// TODO: I'm not sure if this is necessary. Why not just create channel tabs in the specified
     /// order, in TUI?
-    chans: Vec<(String, HashSet<String>)>,
+    chans: Vec<Chan>,
```
trevarj, line 91: I thought this would be a nice cleanup and allow me to hold the number of retries available per channel.

libtiny_client/src/state.rs
```diff
@@ -232,6 +271,31 @@ impl StateInner {
...
+            Reply { num: 477, params } => {
+                // Only try to automatically rejoin if nickserv_ident is configured
+                if self.nickserv_ident.is_some() {
+                    // Get channel name
+                    let channel = params.get(1).unwrap().clone();
+                    // Find channel in self.chans
+                    let idx = utils::find_idx(&self.chans, |c| c.name == channel).unwrap();
+                    // Retry joining channel if retries are available
+                    if self.chans[idx].retry_join() {
+                        let mut snd_irc_msg = snd_irc_msg.clone();
```
trevarj, line 284: I hate having to clone here...I don't think there is any other way, right?

osa1: This is how you do it, yes. It shouldn't be a problem (it should be cheap) as this type is supposed to be cloned and shared with threads/tasks.

libtiny_client/src/state.rs
```diff
@@ -109,13 +109,51 @@ struct StateInner {
...
+        }
+    }
+
+    fn with_nicks(name: String, nicks: HashSet<String>) -> Chan {
+        Chan {
+            name,
+            nicks,
+            retries: 3,
+        }
+    }
```
osa1, line 137: I'm surprised that rustfmt doesn't add an empty line after this line...

libtiny_client/src/state.rs
```diff
@@ -232,6 +271,31 @@ impl StateInner {
                 }
             }
 
+            // Reply 477 when user needs to be identified with NickServ to join a channel
+            Reply { num: 477, params } => {
+                // Only try to automatically rejoin if nickserv_ident is configured
+                if self.nickserv_ident.is_some() {
+                    // Get channel name
+                    let channel = params.get(1).unwrap().clone();
```
osa1, line 279: This unwrap can potentially cause a crash on non-standard-confirming servers (probably on irc.gitter.im for example). Instead we should pattern match on `params.get(1)`, log it if it's `None`.

libtiny_client/src/state.rs
```diff
@@ -232,6 +271,31 @@ impl StateInner {
                 }
             }
 
+            // Reply 477 when user needs to be identified with NickServ to join a channel
+            Reply { num: 477, params } => {
```
osa1, line 275: Could you add a comment here showing the full message format. I had to check and it looks like

```
Msg { pfx: Some(Server("helix.oftc.net")), cmd: Reply { num: 477, params: ["osa1_", "#openjdk", "Cannot join channel (Need to be identified and verified to join this channel, \'/msg NickServ help\' to learn how to register and verify.)"] } }
```

So first arg is our nick, second is channel, then a message.

libtiny_client/src/state.rs
```diff
@@ -232,6 +271,31 @@ impl StateInner {
...
+                // Only try to automatically rejoin if nickserv_ident is configured
+                if self.nickserv_ident.is_some() {
+                    // Get channel name
+                    let channel = params.get(1).unwrap().clone();
+                    // Find channel in self.chans
+                    let idx = utils::find_idx(&self.chans, |c| c.name == channel).unwrap();
```
osa1, line 281: Let's also remove this unwrap which can cause a crash when we reconnect to a server (so the chan tab already exists) but the user closes the chan tab *after* sending JOIN but *before* receiving the JOIN response.

libtiny_client/src/state.rs
```diff
@@ -232,6 +271,31 @@ impl StateInner {
...
+                    // Retry joining channel if retries are available
+                    if self.chans[idx].retry_join() {
+                        let mut snd_irc_msg = snd_irc_msg.clone();
+                        // Spawn task and delay rejoin to give NickServ time to identify nick
+                        tokio::task::spawn_local(async move {
+                            tokio::time::delay_for(tokio::time::Duration::from(
```
osa1, line 287: This `Duration::from` is redundant as the argument is already a `Duration`.

libtiny_client/src/state.rs
```diff
@@ -232,6 +271,31 @@ impl StateInner {
...
+                                tokio::time::Duration::from_secs(10),
+                            ))
+                            .await;
+                            debug!("Attempting to re-join channel {}", channel);
```
osa1, line 291: I wonder if we should generate a client event here (see `snd_ev`) so that we can show the progress in the UI. I think currently the UX is not great: if I just connected to a server where I want to join x, y, and z, and I joined x and y but not z, it's hard to see what's going on with z.

I think 477s should be shown in the channel tabs, so for example currently I see this in server tab:

```
helix.oftc.net: osa1_ #openjdk Cannot join channel (Need to be identified and verified to join this channel, '/msg NickServ help' to learn how to
register and verify.)
```

Instead we should create the channel and show this message there.

Once that's done we can also show the progress in the channel tab, e.g. something like: "Will try to re-join the channel in 10 seconds (attempt 1/3).".

What do you think?

libtiny_client/src/state.rs
trevarj: :woozy_face: Not sure how that happened...

osa1:
Thanks @trevarj ! Just added some inline comments.

libtiny_client/src/state.rs
trevarj: I like this idea. Is this like what you have in mind?: 
image

On the server tab it still shows the same messages.

```
liquid.oftc.net: trevarj_ #vcs-home Cannot join channel (Need to be identified
and verified to join this channel, '/msg NickServ help' to learn how to register
and verify.)
```

trevarj: When the retry attempt is successful, the user will just see MOTD or nothing...

libtiny_client/src/state.rs
```diff
@@ -109,13 +117,76 @@ struct StateInner {
...
+#[derive(Debug)]
+struct Chan {
+    /// Name of the channel
+    name: String,
+    /// Set of nicknames in channel
+    nicks: HashSet<String>,
+    /// Channel joined state
+    join_state: JoinState,
+    /// Join attempts
+    join_attempts: u8,
```
trevarj, line 129: Moved join_attempts out of the Joining variant in JoinState because when the state is Initial there should be join attempts. I thought it's nicer than keeping the value in Initial and in Joining.

libtiny_client/src/state.rs
```diff
@@ -207,21 +278,23 @@ impl StateInner {
...
                             None => {
-                                self.chans.push((chan.to_owned(), HashSet::new()));
+                                self.chans.push(Chan::new(chan.to_owned()));
                             }
                             Some(chan_idx) => {
                                 // This happens because we initialize channel states for channels
                                 // that we will join on connection when the client is first created
-                                self.chans[chan_idx].1.clear();
+                                let chan = &mut self.chans[chan_idx];
+                                chan.join_state = JoinState::Joined;
```
trevarj, line 289: This is the point where the JoinState gets set to Joined -- once the channel is actually joined

libtiny_client/src/state.rs
```diff
@@ -232,6 +305,73 @@ impl StateInner {
...
+                                    MAX_JOIN_RETRIES
+                                );
+                                snd_ev.try_send(create_message(retry_msg)).unwrap();
+                                let snd_irc_msg = snd_irc_msg.clone();
+                                // Spawn task and delay rejoin to give NickServ time to identify nick
+                                let (snd_abort, rcv_abort) = tokio::sync::mpsc::channel::<()>(1);
+                                match &mut chan.join_state {
+                                    JoinState::Initial => chan.set_joining(snd_abort),
+                                    JoinState::Joining { stop_task, .. } => *stop_task = snd_abort,
+                                    JoinState::Joined => {
```
trevarj, line 349: Wasn't really sure how to handle this unexpected state. Also see below where there is another unexpected state.

libtiny_client/src/state.rs
```diff
@@ -512,6 +654,41 @@ impl StateInner {
             }
         }
     }
+
+    /// If channel is in Joining state cancel Joining task, otherwise sent part message
+    fn leave_channel(&mut self, msg_chan: &mut Sender<Cmd>, chan: &str) {
+        if let Some(idx) = utils::find_idx(&self.chans, |c| c.name == chan) {
+            match &mut self.chans[idx].join_state {
+                JoinState::Initial => {
+                    warn!("Unexpected state. Leaving channel before attempting to join.")
```
trevarj, line 663: Another unexpected state.

trevarj:
All simple feedback has been applied. I commented on some code that needs another closer review. Sorry this is taking so long!

osa1 (commented):
Reading this code again, I think the initial state is fine. We can remove it, but if we do it we'll need an additional state for "currently joined channels" that we'll join again on reconnect, and move the `chans` vec to that field in `StateInner::reset`. Perhaps currently solution is fine.

I'll do another pass later.

Currnetly the state transitions are:

- Initial -> Joining: when we get 477 for the channel
- Initial -> Joined: when we were able to join on first attempt
- Joining -> Joined: when we get a JOIN message for the channel
- Joining -> Initial: I think when user does `/connect` while we're trying to join?
- Joined -> Initial: Connection reset
- Joined -> Joining: Is this currently invalid?

I think it'd be good to make sure the last one won't happen, or if it will then explain when.

Would be good to document this in the code also (maybe in the documentation for JoinState).

Btw, one bug I just discovered is if I do `C-c enter` and close tiny while there's a join task we want until the next iteration (~10 seconds) before tiny finally quits.

libtiny_client/src/state.rs
```diff
@@ -109,13 +117,76 @@ struct StateInner {
...
+    /// Channel joined state
+    join_state: JoinState,
+    /// Join attempts
+    join_attempts: u8,
+}
+
+#[derive(Debug)]
+enum JoinState {
+    /// Initial state for Chan
+    Initial,
```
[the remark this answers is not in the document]
trevarj: `NotJoined` seems logical to me

trevarj:
>     * Joining -> Initial: I think when user does `/connect` while we're trying to join?

Right, just a reconnect.
> Joined -> Joining: Is this currently invalid?

Yes, this is unexpected/invalid. 

I will add these state transitions to the doc comment of JoinState.

> I think it'd be good to make sure the last one won't happen, or if it will then explain when.

This match should prevent the bad state. I suppose it doesn't prevent the programmer from doing it somewhere later on, but hopefully the state transition doc will prevent that.

```
match &mut chan.join_state {
                JoinState::NotJoined => chan.set_joining(snd_abort),
                JoinState::Joining { stop_task, .. } => *stop_task = snd_abort,
                JoinState::Joined => {
                error!("Unexpected JoinState for channel.");
                      return;
                }
}
```

libtiny_client/src/lib.rs
```diff
@@ -242,6 +244,8 @@ impl Client {
     /// outgoing messages) will be dropped.
     pub fn quit(&mut self, reason: Option<String>) {
         debug!("quit cmd received");
+        // Kill any retry join tasks to prevent quit delay
+        self.state.kill_join_tasks();
```
trevarj, line 248: Resolves bug where there is a delay on quit from a task running.

libtiny_client/src/state.rs
```diff
@@ -109,13 +121,89 @@ struct StateInner {
...
+    nicks: HashSet<String>,
+    /// Channel joined state
+    join_state: JoinState,
+    /// Join attempts
+    join_attempts: u8,
+}
+
+/// State transitions:
+///    Initial -> Joining: When we get 477 for the channel
+///    Initial -> Joined: When we were able to join on first attempt
```
osa1, line 138:
```suggestion
///    Initial -> Joined: When we get a JOIN message for the channel on first attempt
```

libtiny_client/src/state.rs
```diff
@@ -512,6 +671,50 @@ impl StateInner {
...
+                JoinState::Joined => msg_chan.try_send(Cmd::Msg(wire::part(chan))).unwrap(),
+            }
+        }
+    }
+
+    /// Kills all tasks that are trying to join channels
+    fn kill_join_tasks(&mut self) {
+        for chan in &mut self.chans {
+            if let JoinState::Joining { stop_task } = &mut chan.join_state {
+                stop_task.try_send(()).unwrap();
```
osa1, line 695:
```suggestion
                let _ = stop_task.try_send(());
```

When we call this there will sometimes be a race between the `retry_channel_join` task and this code. For example it's possible that `retry_channel_join` just sent a join message and returned (so the receiving end of the channel is closed) but we haven't heard back from the server yet so we didn't update `JoinState` yet and the sending end is still available. This is a benign race that we can ignore.

libtiny_client/src/state.rs
```diff
@@ -512,6 +671,50 @@ impl StateInner {
...
+    /// If channel is in Joining state cancel Joining task, otherwise sent part message
+    fn leave_channel(&mut self, msg_chan: &mut Sender<Cmd>, chan: &str) {
+        if let Some(idx) = utils::find_idx(&self.chans, |c| c.name == chan) {
+            match &mut self.chans[idx].join_state {
+                JoinState::NotJoined => {
+                    warn!("Unexpected state. Leaving channel before attempting to join.")
+                }
+                JoinState::Joining { stop_task, .. } => {
+                    debug!("Aborting task to retry joining {}", chan);
+                    stop_task.try_send(()).unwrap();
```
osa1, line 684:
```suggestion
                    let _ = stop_task.try_send(());
```

Same as above.

libtiny_client/src/state.rs
```diff
@@ -109,13 +121,89 @@ struct StateInner {
...
+///    Joined -> Initial: Connection reset
+///    Joined -> Joining: Unexpected/Invalid state
+#[derive(Debug)]
+enum JoinState {
+    /// Initial state for Chan
+    NotJoined,
+    /// In the process of joining the channel
+    Joining {
+        /// Sender to kill the retry task if tab is closed
+        stop_task: tokio::sync::mpsc::Sender<()>,
```
osa1, line 150:
```suggestion
        stop_task: Sender<()>,
```

libtiny_client/src/state.rs
```diff
@@ -232,6 +322,73 @@ impl StateInner {
...
+                                let retry_msg = format!(
+                                    "Attempting to rejoin {} in 10 seconds... ({}/{})",
+                                    channel,
+                                    MAX_JOIN_RETRIES - retries,
+                                    MAX_JOIN_RETRIES
+                                );
+                                snd_ev.try_send(create_message(retry_msg)).unwrap();
+                                let snd_irc_msg = snd_irc_msg.clone();
+                                // Spawn task and delay rejoin to give NickServ time to identify nick
+                                let (snd_abort, rcv_abort) = tokio::sync::mpsc::channel::<()>(1);
```
osa1, line 362:
```suggestion
                                let (snd_abort, rcv_abort) = tokio::sync::mpsc::channel(1);
```

libtiny_client/src/state.rs
osa1: I think this is not so unexpected. It happens when: 

- We already have a tab for a channel (e.g. from before the connection timed out)
- We re-connect to the server and send a JOIN message
- But we haven't heard back from the server yet (i.e. no JOIN or 477 response yet)
- User closes the tab

I think this is fine, no need for logging.

libtiny_client/src/state.rs
osa1: So basically we got a 477 for a channel that we've already joined. This is weird, yeah. Good idea to log this with ERROR level.

libtiny_client/src/state.rs
```diff
@@ -232,6 +322,73 @@ impl StateInner {
...
+                                    snd_irc_msg,
+                                    rcv_abort,
+                                ));
+                            } else {
+                                // No more retries
+                                let no_retries_msg = format!("Unable to join {}.", channel);
+                                snd_ev.try_send(create_message(no_retries_msg)).unwrap();
+                            }
+                        } else {
+                            warn!("Could not find channel in server state channel list.");
```
osa1, line 382: :+1:

libtiny_client/src/state.rs
```diff
@@ -232,6 +322,73 @@ impl StateInner {
...
+                        }
+                    } else {
+                        warn!("Received 477 reply but nickserv_ident is not configured.");
```
osa1, line 385: I think this is fine, let's make this `debug!` maybe. Note that 477 does not always mean we need to register. Even if it does, it's perfectly fine for a user to try to join to a +R channel without being registered. It's not a problem with tiny that may need debugging later.

libtiny_client/src/state.rs
```diff
@@ -232,6 +322,73 @@ impl StateInner {
...
+                    }
+                } else {
+                    warn!("Received 477 reply but could not parse parameters");
```
osa1, line 388: Let's show the message format here:

```suggestion
                    warn!("Could not parse 477 reply: {:?}", cmd);
```

osa1 (approved):
Just did another pass.. This is looking quite good now. Thanks again, @trevarj !

libtiny_client/src/state.rs
```diff
@@ -207,21 +293,23 @@ impl StateInner {
                         self.usermask = Some(usermask);
 
                         // Initialize channel state
-                        match utils::find_idx(&self.chans, |(s, _)| s == chan) {
+                        match utils::find_idx(&self.chans, |c| &c.name == chan) {
                             None => {
-                                self.chans.push((chan.to_owned(), HashSet::new()));
+                                self.chans.push(Chan::new(chan.to_owned()));
```
osa1, line 298: This initializes the Chan state with NotJoined instead of Joined, which is a bug probably.

trevarj: I didn't understand this part. I'm assuming that when it's not found / `None` then the channel is not actually joined. Possibly an unexpected state?

osa1: I can think of at least two cases where this can happen:

- User sends a `JOIN ...` message in the server tab, instead of using `/join`.
- The server for whatever reason adds the user to a channel. I think this happened to me before, though I don't remember the details.

I think both branches of this `match` should initialize the `Chan` state as `Joined`.

libtiny_wire/src/lib.rs
```diff
@@ -270,11 +270,7 @@ fn parse_one_message(mut msg: &str) -> Result<Msg, String> {
     let params: Vec<&str> = parse_params(msg);
     let cmd = match msg_ty {
         MsgType::Cmd("PRIVMSG" | "NOTICE") if params.len() == 2 => {
-            let is_notice = if let MsgType::Cmd("NOTICE") = msg_ty {
-                true
-            } else {
-                false
-            };
+            let is_notice = matches!(msg_ty, MsgType::Cmd("NOTICE"));
```
trevarj, line 273: Fixed this random clippy warning

libtiny_client/src/state.rs
trevarj: Done!

PR #244, opened 2020-09-04
Fallback to prefix for server name parsing of RPL_YOURHOST reply

### LINKED ISSUE #239: Don't rely on RPL_YOURHOST parsing for connectivity checking
I discovered this while reviewing #237 (see [my comment][1]). We currently use the server name when sending PINGs to a server to check connectivity, but we can't do that if we don't know the server name, which we currently parse from the RPL_YOURHOST message. When parsing fails `servername` stays `None`, and `send_ping` just returns:

https://github.com/osa1/tiny/blob/efcb30310e0113c51e27cb21bbbd8e6783f064a7/libtiny_client/src/state.rs#L29-L31

This breaks connectivity checks when we can't parse RPL_YOURHOST for some reason. One example of this happening is when connecting to [irc.gitter.im][2]. Gitter's IRC server is not following the standards too closely and caused problems before (#214, #211). Currently RPL_YOURHOST from Gitter looks like this:

```
Msg { pfx: Some(Server("irc.gitter.im")), cmd: Reply { num: 2, params: ["osa1", " 1.10.0"] } }
```

This is quite different than how RFC 2812 defines it:

```
       002    RPL_YOURHOST
              "Your host is <servername>, running version <ver>"
```

**What to do when we can't parse RPL_YOURHOST?** I see a few options:

- Send an empty PING message. The error response will reset the connectivity timer. The problem with this is that we'll see error responses in the server tab. Example: `helix.oftc.net: osa1 PING Not enough parameters`
- Send a PING to yourself. This won't work if we're stuck in nick selection (not sure if this can happen). In that case we can fall back to (1). I just tried this on OFTC and it seems to work.
- Send a HELP message to the server, which doesn't require arguments so will always work. (Note: HELP is ignored by irc.gitter.im)
- (Other ideas?)

Note that (1) and (2) works on irc.gitter.im, which doesn't seem care about the PING argument at all and always responds with a PONG immediately. (3) doesn't work as it doesn't seem to support HELP message.

cc @trevarj 

[1]: https://github.com/osa1/tiny/pull/237#discussion_r471390983
[2]: https://irc.gitter.im/

trevarj:
Falling back to getting the server name from the prefix, since some IRC servers don't comply with the standard. 

Recommended by https://www.irc.com/dev/docs/refs/numerics/002.html

> The \<servername\> may also include other information depending on the server software, for example irc.example.com[203.0.113.15/6667], so we instead recommend relying on the prefix or the RPL_MYINFO (004) numeric if knowing the current server hostname is desired.

Closes #239

### DIFF
--- a/libtiny_client/src/state.rs
+++ b/libtiny_client/src/state.rs
@@ -480,7 +480,7 @@ impl StateInner {
 
                 // An example <servername>: cherryh.freenode.net[149.56.134.238/8001]
 
-                match parse_servername(pfx, params) {
+                match parse_servername(pfx.as_ref(), params) {
                     None => {
                         error!("Could not parse server name in 002 RPL_YOURHOST message.");
                     }
@@ -736,7 +736,7 @@ fn parse_yourhost_msg(params: &[String]) -> Option<String> {
 }
 
 /// Parse the server name from prefix
-fn parse_server_pfx(pfx: &Option<Pfx>) -> Option<String> {
+fn parse_server_pfx(pfx: Option<&Pfx>) -> Option<String> {
     if let Some(Pfx::Server(server_name)) = pfx {
         Some(server_name.to_owned())
     } else {
@@ -746,7 +746,7 @@ fn parse_server_pfx(pfx: &Option<Pfx>) -> Option<String> {
 
 /// Parse server name from RPL_YOURHOST reply
 /// or fallback to using the server name inside Pfx::Server
-fn parse_servername(pfx: &Option<Pfx>, params: &[String]) -> Option<String> {
+fn parse_servername(pfx: Option<&Pfx>, params: &[String]) -> Option<String> {
     parse_yourhost_msg(&params).or_else(|| parse_server_pfx(pfx))
 }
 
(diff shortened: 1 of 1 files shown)


libtiny_client/src/state.rs
```diff
@@ -735,35 +735,57 @@ fn parse_servername(params: &[String]) -> Option<String> {
...
+
+/// Parse server name from RPL_YOURHOST reply
+/// or fallback to using the server name inside Pfx::Server
+fn parse_servername(msg: &Msg) -> Option<String> {
+    use wire::Cmd::Reply;
+    if let Reply { params, .. } = &msg.cmd {
+        parse_yourhost_msg(&params).or(parse_server_pfx(msg.pfx.as_ref()))
+    } else {
+        parse_server_pfx(msg.pfx.as_ref())
+    }
```
osa1, line 755: In the call site we already know that this is a reply, with code 002, so this branch will never be taken. We can remove it completely if you pass `params` and `pfx` to `parse_servername`, instead of `msg`. Something like:

```
fn parse_servername(pfx, params) -> Option<String> {
    parse_yourhost_msg(&params).or_else(|| parse_server_pfx(pfx))
}
```

Note that I used `or_else` instead of `or`, if you use `or` you'll evaluate the argument even when `parse_yourhost_msg(...)` will return `Some(...)`. You can evaluate it lazily by using `or_else` and a closure.

osa1:
Thanks @trevarj . I added two more tests. Will merge when CI finishes.

PR #249, opened 2020-09-11
Improve prefix parsing and ambiguous prefix handling

### LINKED ISSUE #247: Prefix without '!' assumed to be a server
RFC2812 defines prefix as follows:
```
prefix     =  servername / ( nickname [ [ "!" user ] "@" host ] )
```

By this syntax, a prefix can be proven to be a nick if it contains either '!' or '@'. Otherwise, it is ambigious.

However, libtiny_wire treats any prefix not containing a '!' as a server name, and tiny replaces server names with the connection connection server. This leads to situations where all history-playback from a bouncer like soju ends up replacing all nicks with the server name.

For non-zero prefixes, a simple `struct { name, user, host }` would probably make the most sense, removing the distinction. Empty prefixes could then be communicated separately, either through an option or an empty Server enum value, for when tiny needs to add the server name itself in the UI.

osa1:
Previously when we see an ambiguous prefix (e.g. "localhost") we'd assume
server. We now have a prefix variant `Ambiguous` for such cases and resolving
the ambiguity is left to the caller.

When there's an ambiguity we try to do the reasonable thing depending on the
type of the message:

- When the message is a numeric reply we assume server.
- In all other cases (e.g. in JOIN/PART/QUIT/PRIVMSG) we assume nick.

Other changes:

- Removed wildcard patterns to make the code future-proof.
- Refactored some logging and comments.

Fixes #247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
   syntax) are now printed to `tiny_debug_logs.txt` file in the log directory. If
   logging is disabled then the file is created at tiny's working directory.
   (#238)
+- Handling of IRC messages with ambiguous prefix (when it's unclear whether the
+  sender is a server or nick) improved. tiny should now work better with some
+  bouncers such as soju. (#249)
 
 # 2020/06/28: 0.6.0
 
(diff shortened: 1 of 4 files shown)


osa1:
@kennylevinsen  could you try this branch and let me know if it fixes #247?

osa1:
We should add a new variant `Ambiguous` in `Pfx` and leave resolving the ambiguity to the users.

PR #259, opened 2020-10-28
Bump tokio to 0.3.3 (from 0.2)

osa1:
Not ready, tokio-tls still uses old tokio

### DIFF
--- a/term_input/src/lib.rs
+++ b/term_input/src/lib.rs
@@ -152,7 +152,7 @@ impl Input {
     /// Create a new input handler. Requires a `tokio` reactor to be running. Make sure to either
     /// set `stdin` to non-blocking mode or enable non-canonical input, otherwise the `Stream`
     /// implementation will not work. If you are using this with `termbox` it does the latter so no
-    /// need to do anything. See module documentation for link on how to enable non-canonical
+    /// need to do anything. See module documentation for links on how to enable non-canonical
     /// input.
     pub fn new() -> Input {
         Input {
--- a/term_input/src/lib.rs
+++ b/term_input/src/lib.rs
@@ -2,8 +2,14 @@
 
 //! Interprets the terminal events we care about (keyboard input).
 //!
-//! Keyboard events are read from stdin. We look for byte strings of key combinations that we care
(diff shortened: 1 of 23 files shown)


osa1:
This now builds fine, but it hangs.

osa1:
There seems to be a bug in term_input or its integration in libtiny_tui. After the first two keypresses the `Stream` for keys never yields again. Or maybe it's a tokio bug?

osa1:
I asked about the input handling problem in tokio Discord channel and the suggestion came up was to set stdin file descriptor to non-blocking mode manually.

osa1:
I just realized examples in libtiny_tui also have the same problem. Example in term_input work fine though.

osa1:
> I asked about the input handling problem in tokio Discord channel and the suggestion came up was to set stdin file descriptor to non-blocking mode manually.

I tried this today and it didn't make any difference.

term_input/src/lib.rs
```diff
@@ -180,15 +189,16 @@ impl Stream for Input {
         }
 
         // Otherwise read stdin and loop if successful
-        match self_.stdin.poll_read_ready(cx, mio::Ready::readable()) {
-            Poll::Ready(Ok(_)) => {
+        match self_.stdin.poll_read_ready(cx) {
+            Poll::Ready(Ok(mut ready)) => {
+                ready.clear_ready();
```
osa1: I'm not sure if this is right. Perhaps we should move this to the `else` branch below.

term_input/src/lib.rs
```diff
@@ -146,54 +182,36 @@ impl Stream for Input {
...
-                        .stdin
-                        .clear_read_ready(cx, mio::Ready::readable())
-                        .unwrap();
-                    Poll::Pending
+            // Otherwise read stdin and loop if successful
+            let mut poll_ret = self_.stdin.poll_read_ready(cx);
+            loop {
+                match poll_ret {
+                    Poll::Ready(Ok(mut ready)) => {
+                        ready.clear_ready();
```
osa1, line 199: Should I move this to the `else` branch below?

PR #263, opened 2020-11-09
Highlight channels with join/part/quit events. Closed #262

### LINKED ISSUE #262: Highlight channels with join/part/quit events
Currently we have 3 colors/styles for showing a channel name in the tab bar:

- (1) Normal: a faded white/grey
- (2) NewMsg: a pink-ish color, used when a channel gets new messages
- (3) Highlight: bold red, used when a channel has a mention to the user

These are listed form the lowest priority to highest. For example if a channel is shown as (2) and a new message to the user is sent we show it as (3). After that new messages don't turn it back to (2). We reset the state to (1) when the channel gets focus.

I suggest we insert this for (2) and shift (2) and (3):

- (2) JoinOrLeave: yellow, used when channel has join/part/quit activity.

trevarj:
:slightly_smiling_face:

### DIFF
--- a/libtiny_ui/src/lib.rs
+++ b/libtiny_ui/src/lib.rs
@@ -76,6 +76,7 @@ impl MsgSource {
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub enum TabStyle {
     Normal,
+    JoinOrPart,
     NewMsg,
     Highlight,
 }
--- a/tiny/src/conn.rs
+++ b/tiny/src/conn.rs
@@ -239,6 +239,7 @@ fn handle_irc_msg(ui: &dyn UI, client: &Client, msg: wire::Msg) {
                 if ui.user_tab_exists(serv, nick) {
                     ui.add_nick(nick, ts, &MsgTarget::User { serv, nick });
                 }
+                ui.set_tab_style(TabStyle::JoinOrPart, &MsgTarget::Chan { serv, chan: &chan })
             }
         }
 
@@ -260,6 +261,7 @@ fn handle_irc_msg(ui: &dyn UI, client: &Client, msg: wire::Msg) {
                     Some(time::now()),
                     &MsgTarget::Chan { serv, chan: &chan },
                 );
+                ui.set_tab_style(TabStyle::JoinOrPart, &MsgTarget::Chan { serv, chan: &chan })
             }
         }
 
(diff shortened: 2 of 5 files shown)


libtiny_ui/src/lib.rs
```diff
@@ -78,6 +78,7 @@ pub enum TabStyle {
     Normal,
     NewMsg,
     Highlight,
+    JoinOrPart,
```
osa1, line 81: This should be between `Normal` and `NewMsg`, see the documentation above the type:

```
// NOTE: Keep the variants sorted in increasing significance, to avoid updating
// style with higher significance for a less significant style (e.g. updating
// from `Highlight` to `NewMsg` in `set_tab_style`).
```

osa1 (approved):
Welcome back @trevarj ! Looks great! I added a comment.

libtiny_ui/src/lib.rs
trevarj: Oops, I didn't notice it! I will change it tomorrow!

trevarj:
Fixed ordering of enum and rebased

tiny/src/conn.rs
```diff
@@ -239,6 +239,7 @@ fn handle_irc_msg(ui: &dyn UI, client: &Client, msg: wire::Msg) {
                 if ui.user_tab_exists(serv, nick) {
                     ui.add_nick(nick, ts, &MsgTarget::User { serv, nick });
                 }
+                ui.set_tab_style(TabStyle::JoinOrPart, &MsgTarget::Chan { serv, chan: &chan })
```
osa1, line 242: I think this should be in the "else" branch above. Currently we highlight the channel tab even when we join the channel (in the `new_chan_tab` part above), which doesn't make sense. We only want to highlight it when others are joining/leaving.

tiny/src/conn.rs
```diff
@@ -261,6 +262,7 @@ fn handle_irc_msg(ui: &dyn UI, client: &Client, msg: wire::Msg) {
                     &MsgTarget::Chan { serv, chan: &chan },
                 );
             }
+            ui.set_tab_style(TabStyle::JoinOrPart, &MsgTarget::Chan { serv, chan: &chan })
```
osa1, line 265: I think this is not buggy, but I think it would be better to move this to the `if` above, for similar reasons.

tiny/src/conn.rs
trevarj: It is inside the `else` block. Maybe I'm not understanding? Do you mean after line 240?

osa1: Ah, you are right. Sorry for the noise.

osa1:
Perfect, thanks!

PR #266, opened 2020-11-14
Fixed bug with scrollback limit.

### LINKED ISSUE #265: Scrollback limit sometimes causes scrolling too much
It seems like there's currently a bug when removing old lines when scrollback is enabled where I can scroll past the oldest line.

I'm guessing in some cases the same bug would also cause not being able to scroll enough to see the oldest line, but I haven't encountered this case yet.

To reproduce: set the scrollback something small. 100 works fine in my case. Then leave the client running for a while in an active channel. Every once in a while try to scroll to the top using page-up/down. You'll end up with a screen like:

```
---------------
<empty>
<empty>
<empty>
old line
old line
---------------
```

What should happen instead is I should not be able to scroll after the oldest line is at the top:

```
---------------
old line
old line
old line
old line
old line
---------------
```

trevarj:
When a line is removed due to reaching scrollback limit, update the
message area's line height. Added test to assert the height is being
maintained.

Closes #265

### DIFF
--- a/libtiny_tui/src/msg_area/mod.rs
+++ b/libtiny_tui/src/msg_area/mod.rs
@@ -160,9 +160,12 @@ impl MsgArea {
     pub(crate) fn flush_line(&mut self) -> usize {
         let line_height = self.line_buf.rendered_height(self.width);
         // Check if we're about to overflow
+        let mut removed_line_height = 0;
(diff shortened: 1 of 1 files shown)


osa1:
Perfect, thanks @trevarj !

PR #276, opened 2020-11-27
term_input: Enable non-blocking mode

### LINKED ISSUE #269: Freezing on startup on ArchWSL
## Bug
When starting tiny on [ArchWSL](https://github.com/yuk7/ArchWSL) it will instantly freeze on startup no longer responding to any input.
I compiled it myself on the same system and there were no errors.

## Versions
tiny: Commit `4d3532`

osa1:
Normally term_input is used with non-canonical input with VMIN=0 and
VTIME=0. On Linux in this mode stdin does not block (returns 0 when it's
not ready for reading, see [1] and [2]). However on WSL it doesn't work
that way and we have to set stdin to non-blocking mode to be able to do
`read(stdin)` without blocking.

We now set stdin to non-blocking mode when creating an `Input` and
restore the old flags when dropping it.

Fixes #269

[1]: https://www.gnu.org/software/libc/manual/html_node/Canonical-or-Not.html
[2]: https://www.gnu.org/software/libc/manual/html_node/Noncanonical-Input.html

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1041,6 +1041,7 @@ name = "term_input"
 version = "0.2.0"
 dependencies = [
  "libc",
+ "log",
  "mio",
  "nix",
  "term_input_macros",
--- a/term_input/Cargo.toml
+++ b/term_input/Cargo.toml
@@ -9,6 +9,7 @@ edition = "2018"
 
 [dependencies]
 libc = "0.2"
+log = "0.4"
 mio = "0.7"
 nix = "0.19"
 term_input_macros = { path = "../term_input_macros" }
--- a/term_input/src/lib.rs
+++ b/term_input/src/lib.rs
@@ -4,14 +4,6 @@
 //!
 //! Keyboard events are read from `stdin`. We look for byte strings of key combinations that we
 //! care about. E.g. Alt-arrow keys, C-w etc.
-//!
-//! NOTE: Make sure to either set `stdin` to non-blocking mode or enable non-canonical input. For
-//! the latter, see:
-//!
-//! - https://www.gnu.org/software/libc/manual/html_node/Canonical-or-Not.html
-//! - https://www.gnu.org/software/libc/manual/html_node/Noncanonical-Input.html
-//!
-//! Also make sure to set `VMIN` and `VTIME` to zero in non-canonical mode.
 
 #[cfg(test)]
 mod tests;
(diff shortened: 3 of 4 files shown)


osa1:
@LordMZTE could you please try this branch and let me know if this also fixes the issues you're having on your WSL?

LordMZTE:
This did fix the freezing issue, but not the connection issue on both ArchWSL and UbuntuWSL.

osa1:
What are you observing exactly with this patch? What is the error message? What build are you using? (which feature flags?)

LordMZTE:
I am building with no feature flags and the error I'm getting is
```
Resolving host...
19:37 Connection error: failed to lookup address information: Temporary failure in name resolution
```
just as it was before this branch. is there anything I can do to get more output?
I built tiny by using `cargo install --git https://github.com/osa1/tiny.git --branch term_input_stdin_flags`.
If I ping the IRC server, I do get 100% packet loss but the name resolution works. I'm guessing that's normal tho? I doubt that the issue is not being able to connect to the server. After all, it did work with the same config on the pre build binary.

LordMZTE:
These are my openssl versions in case that may cause the issue

## UbuntuWSL
```
❯ apt show libssl-dev
Package: libssl-dev
Version: 1.1.1f-1ubuntu2
<snip>
```

## ArchWSL
```
❯ pacman -Q openssl
openssl 1.1.1.h-1
```

osa1:
> If I ping the IRC server, I do get 100% packet loss but the name resolution works

This is weird, I can ping irc.oftc.net just fine, and get no packet losses:

```
~ $ ping irc.oftc.net
PING irc.geo.oftc.net (130.239.18.116) 56(84) bytes of data.
64 bytes from solenoid.acc.umu.se (130.239.18.116): icmp_seq=1 ttl=45 time=101 ms
64 bytes from solenoid.acc.umu.se (130.239.18.116): icmp_seq=2 ttl=45 time=105 ms
64 bytes from solenoid.acc.umu.se (130.239.18.116): icmp_seq=3 ttl=45 time=98.0 ms
64 bytes from solenoid.acc.umu.se (130.239.18.116): icmp_seq=4 ttl=45 time=100 ms
^C
--- irc.geo.oftc.net ping statistics ---
4 packets transmitted, 4 received, 0% packet loss, time 3004ms
rtt min/avg/max/mdev = 98.008/101.450/105.518/2.765 ms
```

Maybe this is an issue with your firewall or network? I think we can't expect this to work if you can't even ping the server.

Do you get the same error if you try to connect without tls (make sure to update your port number as well, tls and non-tls connections use different port numbers)? In the installation command you're using you build tiny with rustls (Rust implementation of TLS) so your system openssl/libssl version doesn't matter.

osa1:
@LordMZTE let's discuss the network issues in its own issue. Could you try connecting with this settings:

- addr: irc.oftc.net
- port: 6667
- tls: false

Do you get the same error?

I think you should really sort out the ping issue first.

osa1:
I asked about this non-blocking mode issue on SO: https://stackoverflow.com/questions/65043058/linux-non-canonical-input-mode-works-different-in-linux-and-wsl

osa1:
Hm, it seems like this was reported in 2018: https://github.com/microsoft/WSL/issues/3507

LordMZTE:
alright :D i'll create an issue for the connection issues.

Issue #273, opened 2020-11-21
Testing TODOs


osa1:
In this issue I'm listing issues that I'd like to test as a part of the automated tests (`cargo test`):

- #271
- #247 parsing is already tested, but we should also make sure messages without `!` in the prefix are shown in the correct tab.

Issue #278, opened 2020-11-30
PRIVMSG (when none of target or sender is us) is shown with incorrect sender


osa1:
I just got this message in Freenode:

```rust
Msg {
    pfx: Some(
        User { 
            nick: "e",
            user: "e@freenode/staff/spy.edk"
        }
    ),
    cmd: PRIVMSG {
        target: User("$$*"),
        msg: "...",
        is_notice: true,
        ctcp: None
    }
}
```

Shown as:

![Screenshot_2020-11-30_09-40-48](https://user-images.githubusercontent.com/448274/100577120-92b7f000-32f0-11eb-99ca-52a015a68fc2.png)

Here the tab name is right (a PRIVMSG sent to `$$*`, it looks weird but there's no other way to show this message I think) but sender should be `e`, not my nick.

I suspect this is introduced with db38123f4021d994845cc1e8248ddc8be913d90f.

We should first start testing this stuff (#273) before making any more changes to showing PRIVMSGs, as the logic is already complex enough.

osa1:
Relevant bits from the IRC about this `$$*` sytnax:

```
      Command: PRIVMSG
   Parameters: <msgtarget> <text to be sent>

   ...

   The <msgtarget> parameter may also be a host mask (#<mask>) or server
   mask ($<mask>).  In both cases the server will only send the PRIVMSG
   to those who have a server or host matching the mask.  The mask MUST
   have at least 1 (one) "." in it and no wildcards following the last
   ".".  This requirement exists to prevent people sending messages to
   "#*" or "$*", which would broadcast to all users.  Wildcards are the
   '*' and '?'  characters.  This extension to the PRIVMSG command is
   only available to operators.

  msgtarget  =  msgto *( "," msgto )
  ...
  msgto      =/ targetmask
  targetmask =  ( "$" / "#" ) mask
  mask       =  *( nowild / noesc wildone / noesc wildmany )
  wildone    =  %x3F ; "?"
  wildmany   =  %x2A ; "*"
  nowild     =  %x01-29 / %x2B-3E / %x40-FF
                  ; any octet except NUL, "*", "?"
  noesc      =  %x01-5B / %x5D-FF
                  ; any octet except NUL and "\"
```

I don't want to be parsing this, I'm inclined to just show PRIVMSGs with masks in server tabs.

osa1:
[...]

PR #279, opened 2020-11-30
New testing infra and tests

osa1:
This is currently WIP. The goal is to test handling connection events (#273).

We will either have to update the `Client` type to allow creating mock/test clients (similar to `TUI`), or create a `Client` trait and implement a new struct for testing. It was easy to have a single `TUI` struct for testing and actual use, but for clients I think having a trait may be better.


osa1:
This provides a very general way of testing updating the TUI after a connection event. I was thinking maybe there could be a simpler way to test just the PRIVMSG handling, but it's a bit difficult (without significant refactoring), because we need to check whether a tab exists, and highlight the tab differently depending on whether the message mentions us or not etc. these currently need a TUI instance. We could provide a callback for checking whether a tab exists, return an enum for highlighting etc. but I'm not sure if it's worth refactoring existing code that much. Perhaps the current PR may be helpful in testing other stuff as well.

libtiny_tui/src/lib.rs
```diff
@@ -65,6 +64,28 @@ impl TUI {
...
+        let tui = Rc::new(RefCell::new(tui::TUI::new_test(width, height)));
+        let inner = Rc::downgrade(&tui);
+
+        let (snd_ev, rcv_ev) = mpsc::channel(10);
+
+        // We don't need to handle SIGWINCH in testing so the receiver end is not used
+        let (snd_abort, _rcv_abort) = mpsc::channel::<()>(1);
+
+        // Spawn input handler task
+        spawn_local(input_handler(tui, snd_ev, snd_abort));
```
osa1, line 80: This won't work, we have to create a `Stream` for input events (probably a channel so that we can return the sender to the caller and write to it in tests) and pass it to `input_handler`.

osa1:
> We will either have to update the Client type to allow creating mock/test clients (similar to TUI), or create a Client trait and implement a new struct for testing. It was easy to have a single TUI struct for testing and actual use, but for clients I think having a trait may be better.

Perhaps we could refactor the client loop that runs after connecting to a server into its own function and run it in tests.

osa1:
Currently if I run the test with `--nocapture` I see this output:

```










09:02 osa1: blah blah blah

mentions chat.myserver.net $$*
```

Which shows both of the bugs (the tab name and sender) in #278.

I don't understand why there's one line between the tab bar and the message though. It could be another bug or an issue with the test setup.

osa1:
> I don't understand why there's one line between the tab bar and the message though. It could be another bug or an issue with the test setup.

Oh, that's the input line...

PR #285, opened 2020-12-13
Tab area

trevarj:
Each commit message might be useful to look at, but here is a general list of my work so far:

-  Refactored current tab functionality out into a new module `tab_area`. 
- `TUI` now has a `TabArea` which contains the old tab style and the new `TabPanel` style.
- Painful tinkering to get all the input line stuff to use an x offset instead of `0` for its leftmost position.
- `TabPanel` scrolls by page only and uses an up and down arrow to indicate whether there are more pages in either direction.
- A new configuration in `config.yml` - `tui_ui` which can be `line` or `panel`, defaulting to `line` (current functionality).
- A new command to hotswap between tab ui styles  - `/tabui` 🔁 
- Tests for `TabPanel`


trevarj:
TODOs:

- [ ] Solve how to handle copying multiline links (panel gets highlighted) 😢 
- [ ] Scrolling `TabPanel` keybinding
- [ ] Push active tabs to the top of the `TabPanel`

trevarj:
@osa1, channels aligned to the bottom of the tab panel: https://github.com/trevarj/tiny/tree/tab-panel-bottomup

Tall terminal friendly!

trevarj:
image

PR #290, opened 2020-12-21
Removed libtiny-ui

trevarj:
Added TinyUI struct to combine the TUI and Logger. Removes dynamic
dispatch from the UI.

This will allow us to access/configure the Logger from commands.


libtiny_common/src/lib.rs
```diff
@@ -141,3 +141,97 @@ impl Hash for ChanNameRef {
...
+            MsgSource::Chan { chan, .. } => chan.display(),
+            MsgSource::User { nick, .. } => nick,
+        }
+    }
+}
+
+// NOTE: Keep the variants sorted in increasing significance, to avoid updating
+// style with higher significance for a less significant style (e.g. updating
+// from `Highlight` to `NewMsg` in `set_tab_style`).
+// TODO: This shouldn't be a part of the API
```
trevarj, line 212: need to remove this

libtiny_tui/src/lib.rs
```diff
@@ -204,17 +202,17 @@ async fn input_handler<S>(
...
             }
         }
     }
 }
 
-impl UI for TUI {
+impl TUI {
     delegate!(draw());
-    delegate!(new_server_tab(serv_name: &str, alias: Option<String>,));
+    delegate!(new_server_tab(serv_name: &str, alias: &Option<String>,));
```
osa1, line 215: Why did you have to change this?

tiny/src/ui.rs
```diff
@@ -7,16 +7,78 @@ use crate::cmd::{parse_cmd, CmdArgs, ParseCmdResult};
...
+            self.ui.$name( $( $x, )* );
+            if let Some(logger) = &self.logger {
+                logger.$name( $( $x, )* );
+            }
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct TinyUI {
```
osa1, line 29: Let's call this `UI`; it's shorter and there isn't any ambiguity about what's UI is this.

osa1 (commented):
Thanks for doing this. Just a very quick look for now, I'll review in more detail later.

libtiny_tui/src/lib.rs
trevarj: To make this macro happy:
`ui.rs`
```
macro_rules! delegate {
    ( $name:ident ( $( $x:ident: $t:ty, )* )) => {
        pub(crate) fn $name(&self, $($x: $t,)*) {
            self.ui.$name( $( $x, )* );
            if let Some(logger) = &self.logger {
                logger.$name( $( $x, )* );
            }
        }
    }
}
```

Let me know if you can spot a better way. I'm a macro noob

libtiny_tui/src/tui.rs
```diff
@@ -371,7 +370,7 @@ impl TUI {
     }
 
     /// Returns index of the new tab if a new tab is created.
-    pub fn new_server_tab(&mut self, serv: &str, alias: Option<String>) -> Option<usize> {
+    pub fn new_server_tab(&mut self, serv: &str, alias: &Option<String>) -> Option<usize> {
```
osa1, line 373: If you revert this change (and the `clone()` below) you won't have to make the other changed I asked about in the previous review.

libtiny_tui/src/lib.rs
osa1: No need to change the macro, I added another comment below.

tiny/src/main.rs
```diff
@@ -127,20 +127,17 @@ fn run(
...
+                    logger.new_server_tab("mentions", &None);
                     Some(logger)
                 }
             });
 
-        let tui: Box<dyn UI> = match logger {
-            None => Box::new(tui),
-            Some(logger) => Box::new(libtiny_ui::combine(tui, logger)),
-        };
+        let tui = UI::new(tui, logger);
```
osa1, line 135: :+1:

osa1 (approved):
Looks great! I think we could revert the change to `&Option<String>` from `Option<String>` in some of the methods, then the PR should be much smaller and you can remove clonning in a few places. Also this currently doesn't pass the tests. If you could fix those I'll merge it.

Thanks!

libtiny_tui/src/lib.rs
trevarj: @osa1 This is the reason why I changed it:
image

libtiny_logger/src/lib.rs
```diff
@@ -37,25 +36,21 @@ impl Logger {
...
             self.inner.borrow_mut().$name( $( $x, )* )
         }
     }
 }
 
-impl UI for Logger {
-    fn draw(&self) {}
-    delegate!(new_server_tab(serv: &str, alias: Option<String>,));
+impl Logger {
+    pub fn draw(&self) {}
```
osa1, line 46: This could also be removed

PR #292, opened 2020-12-25
First iteration for logging configuration

### LINKED ISSUE #217: [Feature idea] Setting to enable or disable logging per network
It would be nice to be able to enable or disable logging for individual servers. There are multiple servers that are basically just spam of hundreds of users making requests to bots, which don't really need to be logged.

Could be something like:

```
- addr: irc.blah.net
  logging: true/false
…
```
And when `true`, logs would be saved to the `log_dir` as currently done.

trevarj:
The code is a bit rough right now, just wanted to get something done for a first draft.

Things to review/discuss:
- Default logging setting -- right now it defaults to true when `log_dir` is configured. I think this makes sense.
- Command args for `/join` and `/connect`(not connected yet) --  do we need the ability to set server logging, chan logging and user logging? See line 179 of `cmd.rs`
- Do we need a new command for turning logs on/off? i.e `/logs on` | `/logs off`

Closes #217

### DIFF
--- a/libtiny_logger/src/lib.rs
+++ b/libtiny_logger/src/lib.rs
@@ -43,6 +43,17 @@ macro_rules! delegate {
 }
 
 impl Logger {
+    delegate!(add_server_config(
+        serv: &str,
+        enabled: Option<bool>,
+        chans_enabled: Option<bool>,
+        users_enabled: Option<bool>,
+    ));
+    delegate!(add_chan_config(
+        serv: &str,
+        chan: &str,
+        enabled: Option<bool>,
+    ));
     delegate!(new_server_tab(serv: &str,));
     delegate!(close_server_tab(serv: &str,));
     delegate!(new_chan_tab(serv: &str, chan: &ChanNameRef,));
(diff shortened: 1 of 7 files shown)


tiny/src/main.rs
```diff
@@ -125,12 +120,30 @@ fn run(
                     None
                 }
                 Ok(logger) => {
+                    // set logger configs
```
trevarj, line 123: This is the first step. It builds the configuration of the logger with the config file values

libtiny_logger/src/lib.rs
```diff
@@ -77,18 +88,35 @@ impl Logger {
...
-
     /// Maps server names to their fds
     servers: HashMap<String, ServerLogs>,
-
     /// Callback used when reporting errors
     report_err: Box<dyn Fn(String)>,
 }
 
+#[derive(Debug)]
 struct ServerLogs {
```
trevarj, line 98: The inner map of `Logger` now works with `Option<File>` so configs can be stored even when logging is off for a particular server/chan

trevarj:
Added testing (please recommend more cases)
Added "global" / default config -- useful if you want all logging off except for specific channel(s)

osa1:
Thanks for doing this @trevarj . I'll try to review it this week.

PR #293, opened 2020-12-25
Join/part tab style still worked with /ignore enabled

### LINKED ISSUE #291: Tabs still change color for join/part even if /ignore is enabled
Problem: when /ignore is enabled, tabs should not change color on the joining or parting of a user

This is my fault, I think. When I added the `JoinOrPart` tab style I didn't check this scenario.

It comes from these two identical lines in `conn.rs` :
```
ui.set_tab_style(TabStyle::JoinOrPart, &MsgTarget::Chan { serv, chan: &chan })
```
Doesn't look like there's a way to get the ignore from there without adding another function on `ui`, so you could probably change it here in `tui.rs`:
```
    pub(crate) fn set_tab_style(&mut self, style: TabStyle, target: &MsgTarget) {
        self.apply_to_target(target, &|tab: &mut Tab, is_active: bool| {
            if !is_active && tab.style < style {
                tab.set_style(style);
            }
        });
    }
```

trevarj:
Also renamed the function `get_ignore_state` -> `is_showing_status` because it was a bit confusing

Closes #291

### DIFF
--- a/libtiny_tui/src/messaging.rs
+++ b/libtiny_tui/src/messaging.rs
@@ -410,7 +410,7 @@ impl MessagingUI {
         }
     }
 
-    pub(crate) fn get_ignore_state(&self) -> bool {
+    pub(crate) fn is_showing_status(&self) -> bool {
         self.show_status
     }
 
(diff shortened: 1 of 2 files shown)


osa1 (approved):
Thanks! I'll merge this today.

PR #296, opened 2020-12-31
Fixes #295 (backspace didn't work)

### LINKED ISSUE #295: Backspace doesn't work anymore
I pulled the current master (which I hadn't done for several months) and now just realized that I can't delete text via Backspace anymore. This is a bug. It worked in the past, althrough I can't name the exact commit I had previously checked out, which worked.

zserik:
Fixes #295

### DIFF
--- a/term_input/src/lib.rs
+++ b/term_input/src/lib.rs
@@ -107,6 +107,7 @@ byte_seq_parser! {
     [27, 79, 70] => Key::End,
     [27, 91, 52, 126] => Key::End,
     [9] => Key::Tab,
+    [8] => Key::Backspace,
     [127] => Key::Backspace,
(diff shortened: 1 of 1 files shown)


osa1:
I'm surprised backspace worked for you in the past. I just checked commit history of term_input and as far as I can see we never mapped `8` to anything. Could you tell me which terminal you are using so that I can test this and see which commit broke this and why?

zserik:
I'm using `xterm version 351`, installed on Gentoo Linux.

osa1:
It seems to be a change in xterm. If I try with xterm 330 backspace works fine, with 363 it doesn't.

Ideally we should read this stuff from xterm's terminfo, but in the meantime this patch should be good enough.

PR #303, opened 2021-01-18
Fixed unnecessary tab creation on apply to target

### LINKED ISSUE #302: Sending NAMES on a server tab creates new tab for each nick in server
Go to a server tab and send `NAMES` with no parameters, and watch how it opens new tabs for every nick in the response.

I believe it's happening due to the following:

In `conn.rs`:
```
// RPL_NAMREPLY: List of users in a channel
            else if n == 353 && n_params > 3 {
                let chan = &params[2];
                let chan_target = MsgTarget::Chan {
                    serv,
                    chan: ChanNameRef::new(chan),
                };

                for nick in params[3].split_whitespace() {
                    ui.add_nick(wire::drop_nick_prefix(nick), None, &chan_target);  
                }
            }
```

Here is what one RPL_NAMREPLY looks like:
```
[2021-01-16 23:21:17] DEBUG [libtiny_client/src/lib.rs:508] parsed msg: Msg { pfx: Some(Server("barjavel.freenode.net")), cmd: Reply { num: 353, params: ["trevarj_", "=", "#freenode", "Yoda ryzen taubi19_ MylesBorins bigua Harry maxalt anderson fn dfgg deni"] } }
```
Notice the last param (`param[3]`) has a ton of nicks.

trevarj:
A bug appeared when I made the initial fix, where a new User tab is not highlighted due to the call order of `add_privmsg` and `set_tab_style`.

Closes #302

### DIFF
--- a/tiny/src/conn.rs
+++ b/tiny/src/conn.rs
@@ -232,8 +232,8 @@ fn handle_irc_msg(ui: &UI, client: &dyn Client, msg: wire::Msg) {
                                 } else {
                                     MsgTarget::User { serv, nick }
                                 };
-                                ui.set_tab_style(TabStyle::Highlight, &msg_target);
                                 ui.add_privmsg(nick, &msg, ts, &msg_target, false, is_action);
+                                ui.set_tab_style(TabStyle::Highlight, &msg_target);
                             } else {
                                 // PRIVMSG not sent to us. This case can happen in a few cases:
                                 //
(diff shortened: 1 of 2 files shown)


tiny/src/conn.rs
```diff
@@ -232,8 +232,8 @@ fn handle_irc_msg(ui: &UI, client: &dyn Client, msg: wire::Msg) {
                                 } else {
                                     MsgTarget::User { serv, nick }
                                 };
-                                ui.set_tab_style(TabStyle::Highlight, &msg_target);
                                 ui.add_privmsg(nick, &msg, ts, &msg_target, false, is_action);
```
trevarj, line 235: Adds the new tab, then sets the style

osa1:
Thanks @trevarj , and sorry for the delay!

PR #315, opened 2021-04-14
Implement running shell commands for getting passwords

### LINKED ISSUE #246: Implement getting passwords from external commands
Currently users have to store their passwords in the config file as plain text. That makes it difficult to store the config in e.g. git repos, but also storing password in plain text is just not a great idea in general.

In #171 we designed a feature to solve this problem. Basically we allow getting password from external commands. These commands are just strings, except they start with `!` (`!!` is used for escaping the initial `!`). Commands are allowed in the password fields:

- `nickserv_ident` for NickServ identification password
- `pass` for server passwords
- `sasl.password` for SASL authentication password

If I have `! pass my_irc_password` in a `nickserv_ident` field, tiny runs `pass my_irc_password` when the password is needed and uses the output of the command as the password.

(For reading an env variable `! printenv ENV_VAR` can be used)

Parking the TUI while running the command (without blocking the thread) can be implemented as we do it for the editor support.

The main difficulties are:

- What to do in the client while the command is running. The command may ask a master password to the user to unlock the password store, which may take time to enter. Or the user may simply not be in front of the computer when the command is run etc. The client should be handle waiting arbitrarily long while the command is running.

- How to return the command output to the client.

For (1) I think there are two cases that we need to consider:

- When `pass` is a command: to be able to wait arbitrarily long I think we'll have to run the command *before* actually opening a connection to the server, to avoid timeouts while the server is waiting for `PASS`.

- When `nickserv_ident` or `sasl.password` is a command: by the time we need these passwords we already connected to the server. I think we should wait until the command is run and the password is sent to the server before trying to join channels.

  Note that during this wait the client should run as usual and maintain the connection, by responding to PINGs and sending PINGs to check connectivity, as usual.

  (We need to make sure we handle the case where the connection gets lost while waiting for the command output. I think in the design proposed below we don't need to do anything special for this.)

For (2) I propose we add these methods to `Client`:

- `pass_value`: Provide output of the `pass` command to the client.
- `sasl_pass_value`: Same, for SASL password
- `nickserv_pass_value`: Same, for NickServ password

When `pass` field is used, when connecting (after a disconnect, or when connecting for the first time) a client will only try to (re)connect when `pass_value` is called.

When SASL or nickserv_ident is used, after connecting to a server the client will only send `JOIN`s after `sasl_pass_value` or `nickserv_pass_value` is called.

(We should probably have some sanity checks in client that makes sure the `..._value` methods are called at the right times, e.g. it doesn't make sense to call `pass_value` if we're already connected, or `nickserv_pass_value` if we already sent a `JOIN` message. Those cases are probably bugs that are worth reporting.)

cc @trevarj

osa1:
Password fields can now be a map with a `command` key and a string value. The
value is used the shell command to run to get the password.

To keep things simple, all external commands are run on startup.

Fixes #246


crates/tiny/src/config.rs
```diff
@@ -60,34 +62,231 @@ pub(crate) struct Defaults {
     pub(crate) tls: bool,
 }
 
+// TODO FIXME: I don't understand why we need `Default` bound here on `P`
 #[derive(Deserialize)]
-pub(crate) struct Config {
-    pub(crate) servers: Vec<Server>,
+pub(crate) struct Config<P: Default> {
```
osa1, line 67: For some reason I can't remove `P: Default` here if I want `derive(Deserialize)`, even though `Default` impl for `P` should not be necessary for deserialization of this type (see also line 92 below). @trevarj do you know why? I wonder if this is a serde bug.

trevarj: That doesn't seem right. I don't know enough about what serde expects to comment on it though.

This seems to resolve the problem, right?
```suggestion
pub(crate) struct Config<P> {
    #[serde(default)]
    pub(crate) servers: Vec<Server<P>>,
```

trevarj: Could it be that the heuristics it uses does not or cannot correctly find the trait bound? https://serde.rs/attr-bound.html

Casmaine69 (approved):

kmaasrud:
@osa1 What is the status on this PR? I would really like to run tiny without storing my password directly in the config :smile:

crates/tiny/src/config.rs
trevarj: Fixed this by replacing serde's derived trait bounds : https://github.com/trevarj/tiny/commit/b3a22b265564d7609551e8c6c7249fb0275597e3

trevarj: @osa1 let me know if you want me to PR this change, if it's acceptable. Then maybe we can finish this up

osa1: @trevarj it would be great if you could take this over!

osa1:
@kmaasrud I'm a bit low on bandwidth but I have time this week to make some progress. Maybe we can finish this.

Before implementing, there are some design questions to answer though.

> This works fine when the password store is unlocked before running tiny. If not unlocked I think most password managers will want to print something to stdout first, something like Enter master key password: to unlock the password store, which won't work because we capture the stdout, to be able to read the printed password. Not sure how to work around this issue yet.

We have two options:

1. We could run commands before running the TUI.
2. We could use suspend/activate, similar to how we run `$EDITOR`.

I feel like (1) should be good enough, and it will certainly make the implementation simpler.

I think the syntax (`$` prefix) for commands (to distinguish them from passwords) is fine.

crates/tiny/src/config.rs
osa1: @trevarj 's suggestion worked, marking as resolved.

kmaasrud:
> I feel like (1) should be good enough, and it will certainly make the implementation simpler.
> 
> I think the syntax (`$` prefix) for commands (to distinguish them from passwords) is fine.

+1 on this---it sounds like a very good solution to the problem! Better to keep it simple where possible.

osa1:
I think this is done, just need to update README, CHANGELOG, and commit message.

However I'm having trouble using this feature. If I use `pass show irc_pass` as a password command, I get this error when running tiny:

```
Running server password command for server irc.oftc.net (`pass show irc_pass`)
Command returned non-zero: 2
stdout is empty
stderr:
--------------------------------------
gpg: public key decryption failed: No such file or directory
gpg: decryption failed: No such file or directory

--------------------------------------
```

But when I run `pass show irc_pass` in shell it works as expected.

Once I run the `pass` command to unlock the key store (so that `pass show irc_pass` directly prints the password) I can run tiny.

Does anyone know why this happens?

kmaasrud:
> Does anyone know why this happens?

@osa1 Some environment variable that the binary cannot access? Have you tested with different password commands (e.g., a simple echo?)

osa1:
Yeah, simple commands like `nickserv_ident: "$ echo '...'"`, or `cat ...` etc. work fine.

We should figure out how to make this work with `pass`. I'm assuming most people will use some kind of password manager, instead of a simple echo/cat etc.

trevarj:
I just tried and it works for me with the `$ pass show irc-pass` in the `password` field.

The keyring unlock pops up and I type my unlock password and then it starts `tiny`.

kmaasrud:
Using `security` on MacOS also works for me! The keychain unlocks nicely and tiny successfully uses the requested password.

osa1:
I just tried this on my Linux system and it also works there.

I will merge this today, we can debug any platform-specific issues later.

README.md
```diff
@@ -108,23 +108,21 @@ join require identification. To use this method enter your nick password to the
 
 ### Using external commands for passwords
 
-When a password field in the config file starts with `$`, tiny uses rest of the
-field as the shell command to run to get the password.
+When a password field in the config file is a map with a `cmd` key, the value
```
trevarj, line 111: Should we update the sample `config.yml` too?

osa1: We show an example in the README so I'm not sure if it's necessary.

kmaasrud:
Awesome work, @osa1! Thanks for fixing this so quickly :smile:

Issue #79, opened 2018-01-06
Add some indicator for new messages since last time a tab is visited


osa1:
This is something that I miss from my konversation days, and slack also does this. Show some kind of indicator for new messages since the last time the tab was visited. We should probably update this indicator after spending some time in a tab (e.g. 5 seconds) to avoid accidentally updating it (e.g. to avoid updating it when switching through tabs quickly).

(in konversation this is hidden behind a setting: Configure > Chat Window > Marker Lines, enable "Mark the last position in a chat window when it's hidden")

PR #317, opened 2021-04-17
Add configuration to hide mentions tab

### LINKED ISSUE #311: Is it possible to disable the mentions tab?
I never miss a mention, and that special highlighted tab just distracts me :)

Is it possible somehow to disable it?

trevarj:
Closes #311

### DIFF
--- a/crates/tiny/src/main.rs
+++ b/crates/tiny/src/main.rs
@@ -104,25 +104,25 @@ fn run(
     local.block_on(&runtime, async move {
         // Create TUI task
         let (tui, rcv_tui_ev) = TUI::run(config_path.clone());
-        tui.draw();
 
         // Create logger
         let report_logger_error = {
             let tui_clone = tui.clone();
             Box::new(move |err: String| {
-                // Somehwat hacky -- only tab we have is "mentions" so we show the error there
+                tui_clone.new_server_tab("logger", None);
                 tui_clone.add_client_err_msg(
                     &format!("Logger error: {}", err),
-                    &MsgTarget::Server { serv: "mentions" },
+                    &MsgTarget::Server { serv: "logger" },
                 )
             })
         };
         let logger: Option<Logger> =
             log_dir.and_then(|log_dir| match Logger::new(log_dir, report_logger_error) {
                 Err(LoggerInitError::CouldNotCreateDir { dir_path, err }) => {
+                    tui.new_server_tab("logger", None);
                     tui.add_client_err_msg(
                         &format!("Could not create log directory {:?}: {}", dir_path, err),
-                        &MsgTarget::Server { serv: "mentions" },
+                        &MsgTarget::Server { serv: "logger" },
                     );
                     tui.draw();
                     None
@@ -130,6 +130,7 @@ fn run(
                 Ok(logger) => {
                     // Create "mentions" log file manually -- the tab is already created in the TUI so
                     // we won't be creating a "mentions" file in the logger without this.
+                    // If the mentions tab is hidden, we will still log mentions to a file
                     logger.new_server_tab("mentions");
                     Some(logger)
                 }
@@ -160,6 +161,9 @@ fn run(
                     password: auth.password,
                 }),
             };
+
+            // draw tui after server tab is added
+            tui.draw();
 
             let (client, rcv_conn_ev) = Client::new(server_info);
 
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -21,6 +21,9 @@ pub(crate) struct Config {
 
     #[serde(default = "default_max_nick_length")]
     pub(crate) max_nick_length: usize,
+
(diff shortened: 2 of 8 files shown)


crates/tiny/src/main.rs
```diff
@@ -104,25 +104,26 @@ fn run(
     local.block_on(&runtime, async move {
         // Create TUI task
         let (tui, rcv_tui_ev) = TUI::run(config_path.clone());
-        tui.draw();
 
         // Create logger
         let report_logger_error = {
             let tui_clone = tui.clone();
             Box::new(move |err: String| {
                 // Somehwat hacky -- only tab we have is "mentions" so we show the error there
```
trevarj, line 112: This comment can be removed, since now an "errors" tab should pop up, but...

@osa1 do you know how to force this state so I can test it?

osa1: @trevarj you need to somehow return the error value here: https://github.com/osa1/tiny/blob/be66a8026314bbcce98d939401a79c42d279dc56/crates/libtiny_logger/src/lib.rs#L140-L147

Not sure how to add a unit test for this, but for development purposes I think you can modify that code to return an error always.

trevarj: @osa1 what are your thoughts on the UX of this? The "errors" tab will only pop up if there is an issue with the logger (ex. setting a bad path). If you close it then `tiny` shuts down. We can make it so that it can't be closed, then the user will need to fix the error to make it go away.

My opinion is to name the tab "logger" and then also make it not able to close (with /close)

osa1: From UX point of view I think it should be possible to close the "errors" tab.

Here's an idea that I think would work great (though it may be difficult to implement): when all tabs closed (or when no tabs are crated when you run tiny) we will show the "errors" tab, with an input line (so that `/connect` etc. will work), but only if there is at least one error. If the user closes it, then it won't appear again until there's a new error. When it appears, it will appear as the first tab.

Also, if there is only the "errors" tab in the TUI then we may not want to show it in the tab bar.

More concretely:

- If I run the tiny but no server tabs are created, this is what I will see:
  ```
  |                        | // messaging area
  |█                       | // input line
  |                        | // tabs
  ```
  Note that tab line is empty even though we're in `errors` tab

- In case of an error:
  ```
  |Can't create logger: ...| // messaging area
  |█                       | // input line
  |                        | // tabs
  ```
  Ditto, tab line is empty

- When a new tab is created:
  ```
  |Can't create logger: ...| // messaging area
  |█                       | // input line
  |errors irc.oftc.net     | // tabs
  ```
  Note that tab line is empty

- After `/close` in `errors` tab:
  ```
  |...                     | // messaging area
  |█                       | // input line
  |irc.oftc.net            | // tabs
  ```

- After `/close` in `irc.oftc.net` tiny we exit.

@trevarj What do you think about this design in terms of UX and implementation complexity?

trevarj: I like the idea of the blank tab, but it's definitely more changes to add a special case for hiding the tab in the tab line, I think. Might not be worth the mess, but I can look into it.


What would be *really* cool is if we could somehow make it like a welcome screen like vim/emacs. Maybe have it list some of the main commands (/connect, /close, etc). 

```
          ______    _                  
         /_  __/   (_)   ____    __  __
          / /     / /   / __ \  / / / /
         / /     / /   / / / / / /_/ / 
        /_/     /_/   /_/ /_/  \__, /  
                              /____/ IRC
             Basic commands:
             /connect - blah blah
             /join - asdasd
             /close - closes tab

Error: something went wrong...
█ 
welcome irc.oftc.net
```

trevarj:
Welcome screen, always centered (on start):

image
image

trevarj:
For ascii logo: 
https://www.colorschemer.com/ascii-art-generator/?csrfmiddlewaretoken=Dp7XBWwC10PSdNvDyaWzfpvLVSTVtgO8q0onarfbO25Vlgynz0ZFGE0aDn4nLRRq&inputText=Tiny#p=testall&h=0&v=0&f=Speed&t=Tiny

trevarj:
After using this for a while a problem arose:
1. If I am on a channel tab, #tiny and hide the terminal window
2. Someone mentions me
3. The channel floods with more conversation
4. I come back to the terminal window and do not see the mention text (red highlight), unless I scroll up to check manually
I will miss this mention, so I believe that we will have to implement something else to signify that there was a mention to you in the current tab. It may be as simple as highlighting the tab even if you are currently on it. I think this is somewhat related to #79

osa1:
> It may be as simple as highlighting the tab even if you are currently on it

I think this makes sense. The tab would be in "highlight" mode until you see the messages. This is what most other messaging apps work, I think (e.g. slack).

Though perhaps that kind of thing would also require a "jump to next unread message" command?

trevarj:
> "jump to next unread message" command

This would be useful and solve the issue IMO

PR #320, opened 2021-05-05
Recalculate scroll value on window resize

### LINKED ISSUE #319: Invalid scroll value in compact layout after resizing
Screenshot_2021-05-03_10-32-08

In the screenshot above I'm in a channel full of messages, but it shows empty. The reason is because I made the window very small, and scrolled to the top. Then I made the window large again, but the scroll value is not updated, so even though I have less lines to render now, say, 100, the scroll value is something like 1000. As a result the tab doesn't show any messages.

The repro instructions is similar to #318, except after making the window smaller:

- Scroll to the top with `page-up` or `home`
- Make the window larger

You should see the tab empty, as shown in the screenshot above.

cc @trevarj

trevarj:
By maintaining the top-most line index, we can recalculate the scroll
offset on screen resize.

Closes #319

### DIFF
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -48,12 +48,12 @@ impl Layout {
 
 #[derive(Debug, Default)]
 struct Scroll {
-    /// An offset from the last visible line.
-    /// E.g. when this is 0, `self.lines[self.lines.len() - 1]` is drawn at the
-    /// bottom of screen.
+    /// An offset of visible lines from the last visible line of the most recent Line
+    /// E.g. when this is 0, `self.lines.last()` is drawn at the
+    /// bottom of screen. When it is not 0, it is N lines up in the message area.
     scroll: i32,
-    /// Current index of line drawn at the top of the screen
-    /// Used to recalculate `scroll` on resizing of window
+    /// Current index into `self.lines` of the Line drawn at the top of the screen.
+    /// Used to recalculate `scroll` on resizing of window and maintain scrolling in the correct place
     line_idx: usize,
 }
 
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -76,12 +76,10 @@ impl MsgArea {
     }
 
     pub(crate) fn resize(&mut self, width: i32, height: i32) {
-        eprintln!("before {:?}", self.scroll);
         self.width = width;
         self.height = height;
         self.lines_height = None;
         self.lines_height();
-        eprintln!("after {:?}", self.scroll);
     }
 
     pub(crate) fn layout(&self) -> Layout {
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -17,10 +17,8 @@ pub(crate) struct MsgArea {
     width: i32,
     height: i32,
 
-    /// Vertical scroll: An offset from the last visible line.
-    /// E.g. when this is 0, `self.lines[self.lines.len() - 1]` is drawn at the
-    /// bottom of screen.
-    scroll: i32,
+    /// Vertical scroll
+    scroll: Scroll,
 
     line_buf: Line,
 
@@ -48,6 +46,17 @@ impl Layout {
     }
 }
 
+#[derive(Debug, Default)]
+struct Scroll {
+    /// An offset from the last visible line.
+    /// E.g. when this is 0, `self.lines[self.lines.len() - 1]` is drawn at the
+    /// bottom of screen.
+    scroll: i32,
+    /// Current index of line drawn at the top of the screen
+    /// Used to recalculate `scroll` on resizing of window
+    line_idx: usize,
+}
+
 impl MsgArea {
     pub(crate) fn new(width: i32, height: i32, scrollback: usize, layout: Layout) -> MsgArea {
         MsgArea {
@@ -55,7 +64,7 @@ impl MsgArea {
             scrollback,
             width,
             height,
-            scroll: 0,
+            scroll: Scroll::default(),
             line_buf: Line::new(),
             lines_height: Some(0),
             layout,
@@ -67,9 +76,12 @@ impl MsgArea {
     }
 
     pub(crate) fn resize(&mut self, width: i32, height: i32) {
+        eprintln!("before {:?}", self.scroll);
         self.width = width;
         self.height = height;
         self.lines_height = None;
+        self.lines_height();
+        eprintln!("after {:?}", self.scroll);
     }
 
     pub(crate) fn layout(&self) -> Layout {
@@ -87,7 +99,7 @@ impl MsgArea {
         let mut row = pos_y + self.height - 1;
 
         // How many visible lines to skip
-        let mut skip = self.scroll;
+        let mut skip = self.scroll.scroll;
 
         // Draw lines in reverse order
         let mut line_idx = (self.lines.len() as i32) - 1;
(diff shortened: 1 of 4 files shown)


osa1:
Thanks @trevarj ! I Would it be possible to add test(s) for this?

trevarj:
@osa1 Test added, but willing to add more if you can think of a better scenario. Right now the test is as follows:
1. Hit Home to go to first message
2. Resize
3. Assert that the first message is still at the top of the screen

crates/libtiny_tui/src/tests/mod.rs
```diff
@@ -443,6 +443,50 @@ fn test_alignment_long_string() {
...
+    tui.add_privmsg(
+        "osa1",
+        "1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111 1111",
+        ts,
+        &target,
+        false,
+        false,
+    );
+
+    tui.draw();
```
osa1, line 469: Would be good to see the screen before resizing, so that I can compare before and after. Currently I see the after, but I don't know what it was before, so I can't tell immediately if the test is right or wrong, and I have to study it and debug it.

crates/libtiny_tui/src/tests/mod.rs
```diff
@@ -443,6 +443,50 @@ fn test_alignment_long_string() {
...
+    // Should be at the top of message after resize
+    #[rustfmt::skip]
+    let screen =
+        "|00:00 osa1: 1111|
+         |1111 1111 1111  |
+         |1111 1111 1111  |
+         |1111 1111 1111  |
+         |1111 1111 1111  |
+         |                |
+         |< #chan         |";
```
osa1, line 485: What happens if I scroll down in this state?

osa1 (commented):
Thanks @trevarj. I didn't review changes in the TUI yet, but it would be great if we could add a few more tests. Some cases I have in mind:

- Test scroll values without wrapping: if screen height is 10 add 20 short lines. Scroll above so that e.g. lines [5,14] (inclusive) are shown. Now if I resize to add one more line to the screen the lines [5,15] (inclusive) should be shown. (alternatively we should show lines [4,14], but I think adding more lines below, rather than above, to the currently shown range would be better, as new message will be added to the bottom, not top)
- Same as above, but scroll to the bottom so that lines [11,20] are shown. Now resize as before. Last lines should still be shown. (so the range should be [10, 20])
- Same as above, but before and after resizing, add one more message to the tab. The new message should be seen, so the visible range should be updated from [10,20] to [10,21]. After adding a new message try resizing back and forth again. Last line should be visible at all times.

The last line is special because if we scrolled to the bottom (last line visible) then it should stay that way as we resize or as new messages appear, so that we can follow new messages.

So far these cases are all without wrapping. We should also test with wrapping somehow but not sure how..

crates/libtiny_tui/src/tests/mod.rs
trevarj: It will return back down to the bottom, where you will only see the end of the line

crates/libtiny_tui/src/tests/mod.rs
trevarj: No problem, fixing this and adding more tests

trevarj:
@osa1 tests updated. Not sure if I actually captured what you fully described. I tried to comment through each test and show the starting screens.

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -133,8 +146,14 @@ impl MsgArea {
...
-                for line in &mut self.lines {
-                    total_height += line.rendered_height(self.width);
+                let recalculate_scroll = self.recalculate_scroll();
+                for (idx, line) in &mut self.lines.iter_mut().enumerate() {
+                    let height = line.rendered_height(self.width);
+                    // If we need to recalculate, keep adding line heights until we get to line_idx
+                    if recalculate_scroll && idx >= self.scroll.line_idx {
+                        self.scroll.scroll += height;
+                    }
+                    total_height += height;
```
osa1, line 156: It's a bit surprising that a function that is supposed to return total height of a widget re-calculates scroll value. Why is this done here?

I'm also a bit confused about what this is supposed to return. It's not documented, but it's returning `self.lines_height` so I checked the documentation of it, but that also doesn't make sense. It says `total rendered height of all lines` but (1) why would anyone need to know about the height of all lines, only the visible part should be enough (needs documentation) (2) the code uses the scroll value, which isn't relevant if we're calculating the total height of all lines, because that doesn't change depending on the scroll value.

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -169,7 +188,19 @@ impl MsgArea {
...
     pub(crate) fn page_down(&mut self) {
-        self.scroll = max(0, self.scroll - 10);
+        self.scroll.scroll = max(0, self.scroll.scroll - 10);
+    }
+
+    /// Determines if we need to recalculate the scroll offset due to resizing of the window
+    fn recalculate_scroll(&mut self) -> bool {
+        if self.scroll.scroll > 0 {
+            // Subtract current height since our self.scroll.line_idx is the top of the window,
+            // but self.scroll.scroll is the offset from the bottom
```
osa1, line 198: I'm curious, why is this inconsistency exists? Why not make both hold the bottom/offset from bottom or top/offset from top?

osa1 (commented):
Thanks @trevarj . Added some inline comments. Other comments:

- I know this is not added in this PR, but the name `lines_height` is confusing. Should we perhaps rename it to `total_height` or something like that? Any other suggestions?
- Would be good to rename `Scroll` fields to something more helpful, maybe `scroll_from_bottom` and `top_visible_line_idx`?

crates/libtiny_tui/src/tests/resize.rs
```diff
@@ -0,0 +1,267 @@
...
+        |osa1: line13      |
+        |osa1: line14      |
+        |osa1: line15      |
+        |                  |
+        |< #chan           |";
+    expect_screen(screen3, &tui.get_front_buffer(), 18, 11, Location::caller());
+}
+
+#[test]
+fn test_resize() {
```
osa1, line 223: What is this testing exactly? Is it checking that it doesn't crash?

trevarj: Yeah it looks like it tests resizing the window to very small (0x0). I think it was added when we were having crashes like that from all my changes 😅

crates/libtiny_tui/src/msg_area/mod.rs
trevarj: I couldn't sort out how to do the scroll recalculation correctly while keeping the top-most line at the top of the screen during a resize (and also scrolled). You mentioned that it will be good to have it stick there and reveal newer messages when the screen is expanded.

This was easy to use since `self.scroll.line_idx` is set during `draw()` (not nice to have a side effect), where the visible lines are drawn in reverse, from bottom to top. We end on the last visible index (`line_idx`) of `self.lines` and set `self.scroll.line_idx` to that.

crates/libtiny_tui/src/tests/resize.rs
osa1: In that case why not test the whole range of `[(0,0), (N, N)]`? Have a nested loop for (x, y) pairs starting from (0, 0) to (N, N) and resize in the loop body. Code should be smaller too.

crates/libtiny_tui/src/msg_area/mod.rs
trevarj: I have renamed the function and added a doc line to make the purpose more clear.
 > (1) why would anyone need to know about the height of all lines, only the visible part should be enough (needs documentation) 

This is for scrolling, specifically to know what number to set the scroll to

> (2) the code uses the scroll value, which isn't relevant if we're calculating the total height of all lines, because that doesn't change depending on the scroll value.

This is for recalculating the scroll so that we know how to adjust the scroll number during resizing (width changes).

trevarj:
Accidentally delete branch on my fork

PR #322, opened 2021-05-08
Message alignment on draw

### LINKED ISSUE #321: Reloading config does not affect layout
If you change the msg area layout in your config (e.g `layout: aligned` ) and use the command `/reload` it will not change the layout until restarting Tiny.

This can either be removed or fixed. Fixing may help solve #318, which will need the ability to toggle the layout based on window width.

trevarj:
Instead of applying padding to the actual data in the line buffers, this
applies the correct padding on the fly during line height calculation
and drawing.

~~Not a very good solution since I did this in order to be able to switch
layouts on the fly (#321) but it still doesn't work because we don't always
align every line (i.e topic line, server stuff).~~

Closes #321

### DIFF
--- a/crates/libtiny_tui/src/msg_area/line.rs
+++ b/crates/libtiny_tui/src/msg_area/line.rs
@@ -1,9 +1,11 @@
 use crate::line_split::LineType;
+use crate::messaging::{Timestamp, MSG_NICK_SUFFIX_LEN};
 use crate::{
     config::{Colors, Style},
     line_split::LineDataCache,
     utils::translate_irc_control_chars,
 };
+
 use std::mem;
 use termbox_simple::{self, Termbox};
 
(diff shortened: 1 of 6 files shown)


crates/libtiny_tui/src/msg_area/line.rs
```diff
@@ -226,6 +236,58 @@ impl Line {
             }
         }
     }
+
+    /// Checks if padding is needed to align first few segments.
+    /// Returns Vec of modified segments or extra padding segments
+    fn padding(&self) -> Vec<StyledString> {
```
trevarj, line 242: Huge pain in the butt that I could not figure out how to do elegantly 😞

trevarj:
Accidentally deleted branch on my fork

trevarj:
Buggy

PR #323, opened 2021-05-14
Add /quit command

trevarj:
I often see users in the IRC channel wondering how to quit. They usually figure it out, but I think `/quit` is universal among tui IRC clients

### DIFF
--- a/crates/libtiny_tui/src/lib.rs
+++ b/crates/libtiny_tui/src/lib.rs
@@ -181,7 +181,7 @@ async fn input_handler<S>(
                             let result = tui.borrow_mut().try_handle_cmd(&cmd, &from);
                             match result {
                                 CmdResult::Ok => {}
-                                CmdResult::Skip => {
+                                CmdResult::Continue => {
                                     snd_ev.try_send(Event::Cmd { cmd, source: from }).unwrap()
                                 }
                                 CmdResult::Quit => {
--- a/crates/libtiny_tui/src/lib.rs
+++ b/crates/libtiny_tui/src/lib.rs
@@ -22,7 +22,7 @@ mod widget;
 #[cfg(test)]
 mod tests;
 
-use crate::tui::TUIRet;
+use crate::tui::{CmdResult, TUIRet};
 use libtiny_common::{ChanNameRef, Event, MsgTarget, TabStyle};
 
 use futures::select;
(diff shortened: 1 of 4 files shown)


osa1:
Thanks @trevarj .

I think `/quit` shouldn't ask for confirmation. The reason why `C-c` asks for confirmation is because `C-c` is easy to press accidentally, and I think most users would have the muscle memory for it as it's used in shells. If you typed `/quit` you've already made some effort typing all the characters so it's probably not accidental and you really want to quit.

I didn't check other clients. Do they ask for confirmation when you type `/quit`?

trevarj:
@osa1 Good point. I think it should quit automatically too. Not sure about the other clients

trevarj:
Updated to quit immediately.

crates/libtiny_tui/src/tui.rs
```diff
@@ -264,9 +273,10 @@ impl TUI {
                     );
                 }
                 // false to fall through to print help for cmd.rs commands
```
osa1, line 275: Would be good to update this comment as we no longer return `false`.

crates/libtiny_tui/src/tui.rs
```diff
@@ -90,6 +93,12 @@ pub struct TUI {
     config_path: Option<PathBuf>,
 }
 
+pub(crate) enum CmdResult {
+    Ok,
+    Skip,
+    Quit,
+}
```
osa1, line 100: Would be good to document these variants. As far as I understand:

- `Ok`: command is handled by the TUI, don't propagate it to the caller.
- `Skip`: Command is not handled by the TUI, propagate it to the caller.
- `Quit`: Quit requested

If you document it this way then it becomes clear that "skip" is a bit confusing. It feels like the receiver (whoever got the `Skip` value) should skip, but actually it means the sender (whoever created the `Skip` value) is skipped so the receiver should handle it, right? Perhaps it should be called `Ignored` or `Skipped`.

osa1 (approved):
Thanks @trevarj. LGTM. Added some inline comments.

crates/libtiny_tui/src/tui.rs
trevarj: Changed to `Continue`

crates/libtiny_tui/src/lib.rs
```diff
@@ -178,9 +178,17 @@ async fn input_handler<S>(
...
                             // Handle TUI commands, send others to downstream
                             let cmd: String = (&msg[1..]).iter().collect();
-                            let handled = tui.borrow_mut().try_handle_cmd(&cmd, &from);
-                            if !handled {
-                                snd_ev.try_send(Event::Cmd { cmd, source: from }).unwrap();
+                            let result = tui.borrow_mut().try_handle_cmd(&cmd, &from);
+                            match result {
+                                CmdResult::Ok => {}
+                                CmdResult::Continue => {
+                                    snd_ev.try_send(Event::Cmd { cmd, source: from }).unwrap()
```
trevarj, line 185: When we get `Continue`, pass the command down to the client (cmd.rs) to handle it as well.

PR #325, opened 2021-05-17
/msg cmd, first char of nick

### LINKED ISSUE #324: /msg nick pattern error
Reported on IRC:

> using /msg *status znc_help gives a tiny error msg. i think the pattern matched
> for /msg needs to change to support *status

It appears that we don't match on '*' for the nickname in /msg
```
pub(crate) fn is_nick_first_char(c: char) -> bool {
    c.is_alphabetic() || "[]\\`_^{|}".contains(c)
}
```

trevarj:
Changed the /msg command to allow sending to any nick unless it begins
with '#'.

Fixes #324


osa1:
Thanks!

PR #326, opened 2021-05-18
Fix broken commands

trevarj:
This fixes a regression that I made when adding the /quit command.
Accidentally stopped processing commands if it was not in the
tui-specific commands, which causes all tiny client commands (cmd.rs) to
be ignored.


osa1:
Thanks!

PR #328, opened 2021-05-24
Custom key maps

trevarj:
Actions can now be overridden with custom keys

Implementation and usage are fairly straight forward if you look at the `config.yaml`

I won't really be using this, so I tried to make it very simple. Maybe if someone who uses custom key maps has other ideas, then I will add to this. For example, binding a command to a key combo, or adding more combos (`ctrl_shift_x`), etc...

Closes #328


crates/libtiny_tui/src/key_map.rs
```diff
@@ -0,0 +1,278 @@
...
+use serde::de::{MapAccess, Unexpected, Visitor};
+use serde::{Deserialize, Deserializer};
+use term_input::{Arrow, Key};
+
+#[derive(Debug, PartialEq)]
+pub(crate) struct KeyMap(HashMap<Key, KeyAction>);
+
+#[derive(Debug, Copy, Clone, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum KeyAction {
```
trevarj, line 13: A `KeyAction` is dropped in any where a hardcoded `Key` combo was (except in a few areas where it makes no sense, like up arrow for scrolling back in history.

trevarj (commented):
A lot of the diff in `crates/libtiny_tui/src/input_area/mod.rs` is just moving code around

crates/term_input/src/lib.rs
```diff
@@ -110,16 +111,29 @@ byte_seq_parser! {
     [8] => Key::Backspace,
     [127] => Key::Backspace,
     [1] => Key::Ctrl('a'),
+    [2] => Key::Ctrl('b'),
```
trevarj, line 114: Added more supported keys. Did not verify them all, but a few I did verify with `showkey -a`. 
If more keys/combos are useful I can add them too

eoli3n:
### Note to myself

Sandbox
https://gist.github.com/eoli3n/93111f23dbb1233f2f00f460663f99e2

Install
```
pacman -S rust --noconfirm
su - user
cargo install --path crates/tiny
```
Generate config file template
```
~/.cargo/bin/tiny
```

Add to the end of config file
```
key_map:
  ### Defaults ###
  ctrl_c: exit
  ctrl_x: run_editor

  alt_&: 
      tab_goto: 1
  alt_é: 
      tab_goto: 2
  alt_": 
      tab_goto: 3
  alt_': 
      tab_goto: 4
  alt_(: 
      tab_goto: 5
  alt_-: 
      tab_goto: 6
  alt_è: 
      tab_goto: 7
  alt__: 
      tab_goto: 8
  alt_ç: 
      tab_goto: 9
  alt_à: 
      tab_goto: 0
  alt_right:  tab_next
  alt_left:   tab_prev

  shift_up:   messages_scroll_up
  shift_down: messages_scroll_down
  ctrl_u:     input_delete_to_start
  ctrl_d:     exit
  pgup:       messages_page_up
  pgdown:     messages_page_down
  home:       input_move_curs_start
  end:        input_move_curs_end

  left:       input_move_curs_left
  right:      input_move_curs_right
  ctrl_a:     input_move_curs_start
  ctrl_e:     input_move_curs_end
  ctrl_left:  input_move_word_left
  ctrl_right: input_move_word_right
  ctrl_k:     input_delete_to_end
  ctrl_w:     input_delete_prev_word
  backspace:  input_delete_prev_char
  del:        input_delete_next_char
  tab:        input_auto_complete
```

eoli3n:
``input_delete_to_start`` function doesn't exist, would you create it please ?

eoli3n:
Everything is working well with my AZERTY keyboard and my mapping.
Good work, thanks for this @trevarj  !

osa1 (commented):
One concern here is, as far as I can see, we allow keys that we don't support. For example, `C-h` cannot be supported but we allow `ctrl_h`. As a user I would absolutely hate it if I would have to manually debug unsupported keys by making changes in my config, `/reload`, try again...

Maybe we could add a function in `term_input` to return a list of all supported keys, and then have a validation step after parsing to check that the keys user specified are supported.

We should also have a documentation somewhere (perhaps a GitHub wiki page) for the actions, and what they do.

crates/libtiny_tui/src/config.rs
```diff
@@ -17,10 +19,14 @@ pub(crate) struct Config {
     #[serde(default = "usize::max_value")]
     pub(crate) scrollback: usize,
 
+    #[serde(default)]
```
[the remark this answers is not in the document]
trevarj: No, I added it for consistency, but I'm not entirely sure if it matters for `Layout`. For `key_map` it was necessary.

trevarj:
> We should also have a documentation somewhere (perhaps a GitHub wiki page) for the actions, and what they do.

@osa1  Let's enable the wiki and I can write some docs on this and also migrate some stuff from the README over to it for more organization

> Maybe we could add a function in term_input to return a list of all supported keys

Pushing a commit to address this

crates/libtiny_tui/src/config.rs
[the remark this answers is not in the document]
trevarj: Removed in latest commit

crates/libtiny_tui/src/key_map.rs
```diff
@@ -0,0 +1,295 @@
...
+    Disable,
```
osa1, line 14: I think this is not used.

crates/tiny/config.yml
```diff
@@ -139,3 +139,73 @@ colors:
...
+
+
+
+
+
+
+
+
+
+
```
osa1, line 211: Could you remove the whitespace?

crates/libtiny_tui/src/key_map.rs
```diff
@@ -0,0 +1,295 @@
...
+                        "esc" => Key::Esc,
+                        "home" => Key::Home,
+                        "pgdown" => Key::PageDown,
+                        "pgup" => Key::PageUp,
+                        "tab" => Key::Tab,
+                        "up" => Key::Arrow(Arrow::Up),
+                        "down" => Key::Arrow(Arrow::Down),
+                        "left" => Key::Arrow(Arrow::Left),
+                        "right" => Key::Arrow(Arrow::Right),
+                        ch if ch.chars().size_hint().0 == 1 => {
```
osa1, line 217: I think `size_hint` is dangerous in cases like this, quoting the documentation:

> size_hint() is primarily intended to be used for optimizations such as reserving space for the elements of the iterator
> ...
> The default implementation returns (0, None) which is correct for any iterator.

So in principle this could return `0` even though there is one character in the iterator.

For `Chars` generated from a `str` it probably returns the right value, but this kind of generic code often does not survive refactoring and cause bugs in the long term. In this case we could imagine using another iterator instead of `str::chars`. `size_hint` would still work (as in it type checks), but depending on the implementation this code may break.

We should also replace other uses of `size_hint` in the PR, for the same reason.

crates/term_input_macros/src/lib.rs
```diff
@@ -13,10 +17,31 @@ pub fn byte_seq_parser(input: TokenStream) -> TokenStream {
...
+
+fn build_is_valid_key(rules: &[Rule]) -> TokenStream2 {
+    let keys = rules
+        .iter()
+        .map(|r| &r.value.0)
+        .collect::<HashSet<_>>()
+        .into_iter();
+
+    quote!(
+        pub fn is_valid_key(k: Key) -> bool {
```
osa1, line 39: I think name of this function should be derived from `fn_name` above, otherwise we can't use this macro multiple times in a module.

crates/term_input_macros/src/lib.rs
```diff
@@ -13,10 +17,31 @@ pub fn byte_seq_parser(input: TokenStream) -> TokenStream {
...
+            match k {
+                #(#keys => true,)*
+                Key::AltChar(_) | Key::Char(_) => true,
```
osa1, line 42: This case is specific to our use of the macro in `term_input`, so I think it's better to handle it in a function in that module. We could have a function in `term_input` that checks for these cases, then for the default case calls the macro-generated function.

osa1 (commented):
Thanks @trevarj , this is really great. This is requested so many times, it will be great to finally have it. Left a few inline comments.

crates/libtiny_tui/src/key_map.rs
trevarj: It will only be used if someone uses it in their custom config. I added it in case someone wants to disable a keybinding. I can remove if you want

osa1: Ah, OK, that makes sense. Here's a new suggestion then. "disable" is not really a key action, it indicates lack of a key action, so it seems a bit confusing to have it as a variant in key action type. Since we store the mappings in a `HashMap`, why not just skip the entry for a key/key combo if the value is "disable"? I think it could be easily implemented in `Deserialize for KeyMap`.

trevarj:
@osa1 thanks for the review. I addressed all the above comments except for the "Disable" key action

- Fully removed `size_hint`s
- used given #fn_name identifier for macro generated `is_valid_key` check
- moved `term_input` specific logic to non macro-generated function
- removed extra whitespace

crates/libtiny_tui/src/key_map.rs
trevarj: That's probably possible, but it would definitely over complicate the deserialization since it will have to deserialize to `String`, check if `"disabled"`, then convert to `KeyAction`

osa1: This kind of thing is easily implemented using GADTs in languages like Haskell and OCaml, but no such luck in Rust.. Anyways, that's fine. I'll merge this later today.

crates/libtiny_tui/src/input_area/mod.rs
```diff
@@ -333,110 +319,122 @@ impl InputArea {
...
+                    Key::Char(ch) => {
+                        self.modify();
+                        self.buffer.insert(self.cursor as usize, ch);
+                        self.inc_cursor();
+                        WidgetRet::KeyHandled
                     }
-                }
+                    ////////////////////////////////////////////////////////////////////
+                    // Scrolling in history or autocompletion list
+                    Key::Arrow(Arrow::Up) => {
```
osa1, line 332: Why was this not mapped to a `KeyAction`?

crates/term_input_macros/src/lib.rs
```diff
@@ -13,10 +18,31 @@ pub fn byte_seq_parser(input: TokenStream) -> TokenStream {
...
+    quote!(#is_valid_key #byte_parser).into()
+}
+
+fn build_is_valid_key(fn_name: &Ident, rules: &[Rule]) -> TokenStream2 {
+    let keys = rules
+        .iter()
+        .map(|r| &r.value.0)
+        .collect::<HashSet<_>>()
+        .into_iter();
+    let fn_name = format_ident!("{}_is_valid_key", fn_name);
```
osa1, line 38: Didn't know about this macro, nice :+1:

osa1:
I created a Wiki page to list supported key actions and key sequences: https://github.com/osa1/tiny/wiki/Configuring-key-bindings

crates/libtiny_tui/src/input_area/mod.rs
trevarj: Yes, this should be added too. What should the `KeyAction` be called since it covers different modes?

osa1: I think we could treat these modes (next/prev in history or next/prev in autocompletion) as the same. For naming, maybe `NextCompletion` (up) and `PrevCompletion` (down)? Or `NextEntry`/`PrevEntry`?

trevarj: Ah, I see now why I didn't add it. Line 721, `fn autocomplete()` in `input_area/mod.rs` there is a hardcoding of 
```
self.keypressed(Key::Arrow(Arrow::Up), Some(KeyAction::InputAutoComplete));
```

osa1: I'm a bit confused.. I think `InputAutoComplete` is not used anywhere (we don't check for it). Couldn't we remove it?

trevarj: We do check for it:
```
            Some(KeyAction::InputAutoComplete) => {
                if self.exit_dialogue.is_none() {
                    self.input_field.autocomplete(&self.nicks);
                }
                WidgetRet::KeyHandled
            }
```
I will add the `NextEntry/PrevEntry` key actions

osa1: After adding `NextEntry`/`PrevEntry` could we also add one for the characters, and remove the `key` argument from `InputArea::keypressed`? It seems like we won't need it anymore if we also have a case for single characters.

trevarj: Was thinking the same. A `KeyAction::Input(char)` or something?

osa1: That works. We can rename it later if needed.

crates/tiny/config.yml
```diff
@@ -139,3 +139,55 @@ colors:
...
+    # right:      input_move_curs_right
+    # ctrl_a:     input_move_curs_start
+    # ctrl_e:     input_move_curs_end
+    # ctrl_left:  input_move_word_left
+    # ctrl_right: input_move_word_right
+    # ctrl_k:     input_delete_to_end
+    # ctrl_w:     input_delete_prev_word
+    # backspace:  input_delete_prev_char
+    # del:        input_delete_next_char
+    # tab:        input_auto_complete
```
osa1, line 193: Could you add a `'\n` at the end of this line? Without it adding new lines generates extra diffs. We've also always had a newline at the end of the default config so far.

crates/libtiny_tui/src/messaging.rs
```diff
@@ -135,56 +135,48 @@ impl MessagingUI {
         self.msg_area.draw(tb, colors, pos_x, pos_y);
     }
 
-    pub(crate) fn keypressed(&mut self, key: Key) -> WidgetRet {
-        match key {
-            Key::Ctrl('c') => {
+    pub(crate) fn keypressed(&mut self, key_action: Option<KeyAction>) -> WidgetRet {
```
osa1, line 138: This function does not handle `None` case (returns `KeyIgnored`), I think if we make the argument `KeyAction` instead of `Option<KeyAction>` it becomes clear to the caller that this does not handle when a key is not mapped to an action.

If we do that then we decide what to do when a key is not mapped in the code where we try to map the key to an action. It's much nicer, I don't need to follow the value in the code base to find out where the `None` case is handled and how.

Could you implement that? Should be a few lines of change in `TUI::keypressed`.

trevarj: Done!

osa1:
Thanks @trevarj !

097115:
Great work, thanks! But we can't map the functional keys (F1...F12...) right now, am I correct?

trevarj:
@097115 right, those keys aren't captured correctly by Tiny and aren't supported yet. I can look into it and see what it would take to add them.

097115:
@trevarj Would be great, thanks once again!

097115:
@trevarj 

Sorry to disturb you, but a bit of issue: ShiftLeft and ShiftRight keys aren't defined in `crates/term_input/src/lib.rs b/term_input/src/lib.rs`, probably something like this should be added at line 105:

    [27, 91, 49, 59, 50, 68] => Key::ShiftLeft
    [27, 91, 49, 59, 50, 67] => Key::ShiftRight

trevarj:
@097115 Are you sure? The terminal doesn't seem to distinguish between left and right shift when I do `showkey -a`

trevarj:
@097115 thanks. I will add these

Issue #294, opened 2020-12-30
Supporting server commands as / commands


osa1:
micr0 reported this on IRC. We currently don't have / commands for server commands like `WHOIS`. The idea is that you can always send commands directly to a server by sending it in the server's tab. However tab completion doesn't work in the server tab, so if you have a long and weird nick in a channel and want to do `WHOIS` you can't do `/whois ab<tab>`, and in the server tab `WHOISE ab<tab>` won't work. You have to copy the nick and paste it in the server tab.

If we supported `/whois` then `/whois ab<tab>` would complete the nick in the channel tab.

PR #337, opened 2021-07-03
Add support for /whois command

prateeknischal:
Add support for the /whois command using the raw_msg function on the
connection for the particular server.


osa1:
Sorry for missing this @prateeknischal. Given that it's easy to send a raw message to a server (just send the message in the server tab), I'm not sure if adding more code for simple commands like this is worth it. For the `WHOIS` command you can just send `WHOIS nick` in the server tab.

If we start adding commands for simple IRC commands like this, then for completeness we will probably want `LIST`, `RULES`, ... not sure if that's necessary.

WDYT? cc @trevarj

trevarj:
Not the first time this was asked for, due to the ability to autocomplete a username after `/whois ...`. Therefore, I think it's ok to add. It seems that irssi supports all the [commands](https://irssi.org/documentation/help/index_(1.0)/) 🤷🏻‍♂️

prateeknischal:
Hi @osa1 , As @trevarj mentioned, irssi supports a lot of commands and that's why I had the thought of adding it to tiny, to enhance the UX for the client. If this seems redundant, then we can close this PR :)

trevarj:
Note: Relates to #294

osa1:
> ability to autocomplete a username after /whois ....

I think this is a good idea, I'm convinced. However there's a bit of an UX issue with this implementation. If I run `/whois osa1` in channel `#tiny`, I get the response in the server tab. Shouldn't I see it in the same tab that I ran the command in?

trevarj:
> get the response in the server tab

This makes sense to me, since I wouldn't want the WHOIS response to flood the chan tab, even though it is only 6-7 lines.

osa1:
> This makes sense to me, since I wouldn't want the WHOIS response to flood the chan tab, even though it is only 6-7 lines.

I'm not sure. If I run a command in a tab I expect the response in the same tab, unless the command creates a new tab (e.g. `/msg`).

Imagine a new user trying this command. It's possible they won't even realize that the response is in the server tab.

trevarj:
>  It's possible they won't even realize that the response is in the server tab.

Ok, good point. I did not try this yet, but I would expect it to work like doing `/msg chanserv help` in a chan tab --> the server tab actually tags you and the tab name highlights bold. Maybe /whois should work the same as that?

osa1:
> the server tab actually tags you and the tab name highlights bold. Maybe /whois should work the same as that?

Maybe. We could also automatically switch to the server tab. I'd prefer showing the response in the same tab though.

trevarj:
>I'd prefer showing the response in the same tab though.

Ok, in that case, we need to handle the WHOIS replies to direct them to `CurrentTab`:

```
cmd: Reply { num: 311, params: ["tester100", "trev", "~trev", "user/trev", "*", "t"] } }
cmd: Reply { num: 312, params: ["tester100", "trev", "zinc.libera.chat", "Espoo, FI"] } }
cmd: Reply { num: 671, params: ["tester100", "trev", "is using a secure connection"] } }
cmd: Reply { num: 317, params: ["tester100", "trev", "12164", "1629179930", "seconds idle, signon time"] } }
cmd: Reply { num: 330, params: ["tester100", "trev", "trev", "is logged in as"] } }
cmd: Reply { num: 318, params: ["tester100", "trev", "End of /WHOIS list."] } }
```

osa1:
IIUC from the discussion above this needs some more work, but it's been inactive for a about 4 years now. Closing for now.

PR #339, opened 2021-07-04
Added extra key support

### LINKED ISSUE #338: Function-keys (F1-F12) support
It may be possible to capture these keys correctly and be able to bind them to `KeyActions`

> But we can't map the functional keys (F1...F12...) right now, am I correct?

_Originally posted by @097115 in https://github.com/osa1/tiny/issues/328#issuecomment-873522382_

trevarj:
These keys are now available for binding:
- Function keys (F1-F12)
- Ctrl/Alt/Shift + Function keys, ex. `alt_f2`
- Shift + Left/Right Arrow
- Ctrl + \
- Ctrl + ]
- Ctrl + /

Closes #338

### DIFF
--- a/crates/libtiny_tui/src/key_map.rs
+++ b/crates/libtiny_tui/src/key_map.rs
@@ -111,8 +111,8 @@ impl Default for KeyMap {
             (Key::Ctrl('d'), KeyAction::MessagesPageDown),
             (Key::PageUp, KeyAction::MessagesPageUp),
             (Key::PageDown, KeyAction::MessagesPageDown),
-            (Key::ShiftUp, KeyAction::MessagesScrollUp),
-            (Key::ShiftDown, KeyAction::MessagesScrollDown),
+            (Key::Shift(Arrow::Up), KeyAction::MessagesScrollUp),
+            (Key::Shift(Arrow::Down), KeyAction::MessagesScrollDown),
             (Key::Home, KeyAction::MessagesScrollTop),
             (Key::End, KeyAction::MessagesScrollBottom),
             (Key::Tab, KeyAction::InputAutoComplete),
@@ -254,8 +254,8 @@ impl<'de> Deserialize<'de> for MappedKey {
                             ch => Key::Ctrl(single_key(ch)?),
                         },
                         "shift" => match k2 {
-                            "up" => Key::ShiftUp,
-                            "down" => Key::ShiftDown,
+                            "up" => Key::Shift(Arrow::Up),
+                            "down" => Key::Shift(Arrow::Down),
                             unexp => return Err(E::invalid_value(Unexpected::Str(unexp), &Self)),
                         },
                         unexp => return Err(E::invalid_value(Unexpected::Str(unexp), &Self)),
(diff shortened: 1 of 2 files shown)


crates/libtiny_tui/src/key_map.rs
```diff
@@ -254,8 +254,8 @@ impl<'de> Deserialize<'de> for MappedKey {
                             ch => Key::Ctrl(single_key(ch)?),
                         },
                         "shift" => match k2 {
-                            "up" => Key::ShiftUp,
-                            "down" => Key::ShiftDown,
+                            "up" => Key::Shift(Arrow::Up),
+                            "down" => Key::Shift(Arrow::Down),
```
osa1, line 258: I think this should be updated with left/right cases?

osa1:
Looks great, thanks @trevarj . Let me know when you finished parsing and this is ready for merging.

crates/libtiny_tui/src/key_map.rs
trevarj: Yes, sorry. Moved to draft to add more keys (such as Ctrl/Shift/Alt + F Keys).

crates/libtiny_tui/src/key_map.rs
```diff
@@ -232,30 +233,36 @@ impl<'de> Deserialize<'de> for MappedKey {
...
                     Some((k1, k2)) => match k1 {
                         "alt" => match k2 {
-                            "up" => Key::AltArrow(Arrow::Up),
-                            "down" => Key::AltArrow(Arrow::Down),
-                            "left" => Key::AltArrow(Arrow::Left),
-                            "right" => Key::AltArrow(Arrow::Right),
+                            arrow if Arrow::from_str(arrow).is_ok() => {
+                                Key::AltArrow(Arrow::from_str(arrow).unwrap())
+                            }
+                            f if FKey::from_str(f).is_ok() => Key::AltF(FKey::from_str(f).unwrap()),
```
trevarj, line 247: @osa1 can you think of a cleaner way to do these mappings?

osa1: @trevarj We should not have a `FromStr` in `term_input` as parsing is specific to one downstream. For example, you could imagine another user that wants to map `"Up"` instead of `"up"` to `Arrow::Up`, or parse case sensitively. As far as I'm aware currently there's one user for term_input which is tiny, but it's still a good idea to keep tiny specific code out of term_input.

Regarding `is_ok()` and `unwrap()` parts: I think something like this would be fine:

```rust
match v {
    "backspace" => Key::Backspace,
    "del" => Key::Del,
    "end" => Key::End,
    "esc" => Key::Esc,
    "home" => Key::Home,
    "pgdown" => Key::PageDown,
    "pgup" => Key::PageUp,
    "tab" => Key::Tab,
    other => {
        if let Some(arrow) = parse_arrow(other) {
            ...
        } else if let Some(f_key) = parse_f_key(other) {
            ...
        } else {
            ...
        }
    }
}
```

trevarj: @osa1 that's perfect, thanks. Made the change.

PR #340, opened 2021-07-04
Enhanced /help command

### LINKED ISSUE #287: C-? to display keybindings?
Its nice that /help shows all the commands help

How about `C-?` to show all the keybindings?

trevarj:
Entering /help will now spawn a new "dummy" tab that shows all the
possible commands, along with all the currently configured key bindings.

Closes #287


trevarj:
Looking for some feedback/suggestions here.

trevarj:
Here is how the "help" tab currently looks with this patch. You can scroll up and down and /close it.

```
Commands:
/away       - Sets/removes away message - Usage: `/away` or `/away <message>`
/clear      - Clears current tab        - Usage: `/clear`
/close      - Closes current tab        - Usage: `/close`
/connect    - Connects to a server      - Usage: `/connect <host>:<port>` or `/connect` to reconnect
/help       - Displays this message     - Usage: `/help`
/ignore     - Ignore join/quit messages - Usage: `/ignore`
/join       - Joins a channel           - Usage: `/join <chan1>,<chan2>,...` or `/join` in a channel tab to rejoin
/me         - Sends emote message       - Usage: `/me <message>`
/msg        - Sends a message to a user - Usage: `/msg <nick> <message>`
/names      - Shows users in channel    - Usage: `/names`
/nick       - Sets your nick            - Usage: `/nick <nick>`
/notify     - Set channel notifications - Usage: `/notify [off|mentions|messages]`
/quit       - Quit tiny                 - Usage: `/quit`
/reload     - Reloads config file       - Usage: `/reload`
/switch     - Switches to tab           - Usage: `/switch <tab name>`
Key Bindings:
Cancel               - Esc
Exit                 - Ctrl('c')
Exit                 - Ctrl('d')
InputAutoComplete    - Tab
InputDeleteNextChar  - Del
InputDeletePrevChar  - Backspace
InputDeletePrevWord  - Ctrl('w')
InputDeleteToEnd     - Ctrl('k')
InputDeleteToStart   - Ctrl('u')
InputMoveCursEnd     - Ctrl('e')
InputMoveCursEnd     - End
InputMoveCursLeft    - Arrow(Left)
InputMoveCursRight   - Arrow(Right)
InputMoveCursStart   - Ctrl('a')
InputMoveCursStart   - Home
InputMoveWordLeft    - CtrlArrow(Left)
InputMoveWordRight   - CtrlArrow(Right)
InputNextEntry       - Arrow(Down)
InputPrevEntry       - Arrow(Up)
InputSend            - Char('\r')
MessagesPageDown     - PageDown
MessagesPageUp       - PageUp
MessagesScrollDown   - ShiftArrow(Down)
```

crates/libtiny_tui/src/tui.rs
```diff
@@ -626,6 +612,42 @@ impl TUI {
         tab.set_cursor(cursor);
     }
 
+    pub fn show_help_tab(&mut self, messages: &[String]) {
+        let tab_name = "help";
+        if let Some(tab_idx) = self.new_server_tab(tab_name, None) {
+            let target = &MsgTarget::Server { serv: tab_name };
+
+            let mut cmds = vec!["\x0303Commands: ".to_string()];
```
osa1, line 620: Would be good to add a comment here saying that TUI supports IRC color codes and this is and `\x0303` part is IRC color code for ... color (green?).

osa1 (approved):
Great work @trevarj , thanks! I'm surprised by how simple this is.

I think this line needs to be updated: https://github.com/osa1/tiny/blob/8e7a68aaeab59d72b7c285cd3448ec90819ef2b0/crates/tiny/src/cmd.rs#L297

Does this have any problems with showing connection events in help tab? I'm guessing it works exactly the same way as "mentions" tab?

trevarj:
> Does this have any problems with showing connection events in help tab? I'm guessing it works exactly the same way as "mentions" tab?

Yeah, it's identical. I will update that line, thanks for catching it

osa1:
I thought I sent this comment but apparently I didn't.. This looks great but one problem I had is it switches to the help tab, which is fine, but when I close the tab I'm not back in the tab I was before `/help`. This is quite annoying to me as a user. Firefox handles this nicely. If you create a new tab with `C-t` and then close it with `C-w` you're back in the tab before doing `C-t`. It would be very annoying if it instead moved you to the tab left to the new tab created with `C-t`.

@trevarj WDYT?

trevarj:
@osa1 seems like that's a general feature of the tui.

I can store `previous_tab` on the tui and always go back to it when a tab is closed on this PR, or make a new issue and address it. Up to you.

osa1:
Yeah I think storing the previous tab and switching to it when closing a tab would fix the issue. I'm trying to think if it would be useful in any other case. I just tried this in Firefox and it seems to use a similar idea. If I move left and close the tab the tab on the right is selected. If I move right and close the tab the tab on the left is selected. However, if I jump to a specific tab with `alt+<number>` then close it, I'm moved one next to it, instead of the active tab before jumping. I'm not sure if I like this, though I didn't even think about this until now so clearly I don't care too much. We don't have to duplicate this behavior, but I'm just trying to understand what will be the most intuitive.

trevarj:
@osa1 I think it should probably just get logged as another issue since we will need to decide if:
- we want to keep a history stack of all previously visited tabs and maintain it when tabs are closed
- we only want to keep track of the last tab and then fall back to the nearest tab (current behavior)
- or simply don't move to the /help tab 😁 

Any way seems easy to implement

osa1:
@trevarj another idea:

- Make help tab move freely in and out of "groups", for example if my tab bar looks like
  ```
  mentions server1 #channel1 server2 #channel2 help
  ```
  If I move help tab to the left it should look like
  ```
  mentions server1 #channel1 server2 help #channel2
  ```
  One more:
  ```
  mentions server1 #channel1 help server2 #channel2
  ```
  and so on.

- Create help tab next to the current tab, so when I close it I'm back to where I was.

This seems to solve the problem of losing current tab after `/help` and then closing the help tab without adding any extra state in TUI (previous tab).

trevarj:
@osa1 I made "help" a chan tab instead of a "serv" tab so that it opens next to the current tab you're on.

osa1:
It doesn't seem to work correctly. If I do `/help` in a server tab the tab created as the last tab in the group.

Also, I'm not sure if creating as a channel tab a good idea. I can't move it past groups, as mentioned in my comment above.

I wonder if we should make `MsgSource` optional in tabs. "mentions" and "help" tabs doesn't have `MsgSource`s conceptually (they don't belong to a server). If we do that then we can allow freely moving tabs without `MsgSource`s, and possibly remove some hard-coded checks for mentions and help tabs and check whether `MsgSource` is available instead.

crates/libtiny_tui/src/tui.rs
```diff
@@ -1212,7 +1263,7 @@ impl TUI {
 
             MsgTarget::AllServTabs { serv } => {
                 for (tab_idx, tab) in self.tabs.iter().enumerate() {
-                    if tab.src.serv_name() == serv {
+                    if tab.src.serv_name().map_or(false, |s| s == serv) {
```
trevarj, line 1266: If we were on nightly I could use `.contains()` here...

osa1: You could do

```suggestion
                    if tab.src.serv_name() == Some(serv) {
```

trevarj: 😮

crates/libtiny_common/src/lib.rs
```diff
@@ -4,6 +4,10 @@ use std::borrow::Borrow;
 use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 
+pub const MENTIONS_TAB: &str = "mentions";
+pub const HELP_TAB: &str = "help";
+pub const ERRORS_TAB: &str = "errors";
```
osa1, line 9: Interesting idea. I'm not opposed in principle, but would be good to keep this PR as small as possible and move unrelated changes to others as reviewing is already taking too much time even for small changes (sorry :sweat:).

crates/libtiny_common/src/lib.rs
```diff
@@ -157,6 +161,9 @@ pub enum MsgTarget<'a> {
     /// Show the message in the privmsg tab.
     User { serv: &'a str, nick: &'a str },
 
+    /// Show the message in a miscellaneous tab.
+    Misc { name: &'a str },
```
osa1, line 165: Since we know a `Misc` tab is always one of mentions or help (or "errors" though I'm not sure if that should be a part of this PR, see above), I think we can avoid the string here and use a `MiscTab` enum:

```
enum MiscTab {
    Mentions,
    Help,
    ...
}
```

This makes invalid states like `Misc { name: "wat" }` unrepresentable, and comparisons are also more efficient as we avoid checking string equality and check integer equality.

crates/libtiny_logger/src/lib.rs
```diff
@@ -445,6 +445,20 @@ impl LoggerInner {
...
+                // create mentions tab log file like we do with user tabs
+                let mut path = self.log_dir.clone();
+                path.push("mentions.txt");
+                if let Some(mut fd) = try_open_log_file(&path, &*self.report_err) {
+                    report_io_err!(self.report_err, print_header(&mut fd));
+                    f(&mut fd, &*self.report_err);
+                }
+            }
+
+            MsgTarget::Misc { .. } => {
```
osa1, line 459: If you add an enum as suggested above, then we can explicitly match for unhandled cases here and make the code future-proof. If we later add another enum variant this code becomes a compile error so we don't forget to update it.

osa1 (commented):
Thanks @trevarj , this looks great now. I'll make another pass hopefully later today :sweat_smile:

crates/libtiny_common/src/lib.rs
trevarj: @osa1 removed!

trevarj:
@osa1 ok, I think we're really close!! hopefully this is the last commit

crates/libtiny_tui/src/tui.rs
```diff
@@ -1058,7 +1130,9 @@ impl TUI {
                     .splice(insert_idx..insert_idx, to_move.into_iter());
                 self.select_tab(insert_idx);
             }
-        } else if !self.is_server_tab(self.active_idx - 1) {
+        }
+        // Misc tabs get captured in server group
```
osa1, line 1134: Hm.. this is a bit unexpected, but I guess there's no other way to implement it, so it's fine.

crates/tiny/src/ui.rs
```diff
@@ -172,15 +174,18 @@ pub(crate) fn send_msg(
...
         if clients.is_empty() {
             ui.add_client_err_msg(
                 "No connected server found, please use `/connect <server>` to connect to a server",
                 &MsgTarget::CurrentTab,
             );
         } else {
             ui.add_client_err_msg(
-                "You are on the mentions tab, please use `/switch <tab name>` to switch to a tab",
+                &format!(
+                    "You are on the {} tab, please switch to a channel tab to send a message (`/switch <tab name>`)",
```
osa1, line 186: You can send a message in server and user tabs too. I suggest something like

> You are on the {} tab, please switch to a server, channel, or user tab to send a message

I know this is not part of this PR, but I would also remove `/switch ...`. It's often easier to switch using `C-n`/`C-p`, I think suggesting `/switch` but not those is a bit weird. However, the user may have mapped next-tab/prev-tab to another key, so if we want to mention those keys too we should look next-tab/prev-tab in the keymap and suggest accordingly. It's simpler to just not say how to do it.

My approach is: if we can't make it perfect then often it's better to avoid than doing it poorly.

crates/libtiny_tui/src/tui.rs
```diff
@@ -626,6 +643,55 @@ impl TUI {
...
+    pub(crate) fn show_mentions_tab(&mut self) -> Option<usize> {
+        let idx = self.new_misc_tab(MiscTab::Mentions, Some(0));
+        self.add_client_msg(
+            "Any mentions to you will be listed here.",
+            &MsgTarget::Misc(MiscTab::Mentions),
+        );
+        idx
+    }
+
+    pub(crate) fn show_help_tab(&mut self, messages: &[String]) {
```
osa1, line 656: It's worth documenting here that this won't create a new tab and won't update existing tab's contents if a help tab already exists.

Also that this function will automatically select the help tab (when created or otherwise).

crates/libtiny_tui/src/tui.rs
```diff
@@ -525,6 +498,50 @@ impl TUI {
...
+                        Notifier::Mentions
+                    } else {
+                        Notifier::Off
+                    },
+                    None,
+                );
+                Some(tab_idx)
+            }
+            Some(idx) => {
+                self.select_tab(idx);
```
osa1, line 529: This is very confusing. Switching to the help tab happens in two entirely different places.

1. Here when tab already exists
2. In `show_help_tab` when tab doesn't already exist

Need to refactor this somehow. I have a suggestion in my top-level comment.

osa1 (commented):
I found the names `show_mentions_tab` and `show_help_tab` a bit confusing. To me the name "show..." suggests showing something that already exists, but they also create the tab if they don't exist.

I suggest renaming them to "add_help_tab" or "create_help_tab", and then if we do the switching in the command implementation we don't have to explain (document) to the user that "add_help_tab" will also switch to the tab.

Re: switching to the help tab. I suggest *not* switching to a misc tab in the tab creating functions, as in general we may not want to switch to a misc tab on creation always (e.g. I think for "errors" tab it's fine to just highlight it with bold red, no need to switch to it). It's specific to the tab.

So I'd completely separate tab creating and switching logic and do switching in command implementation.

crates/libtiny_tui/examples/bench.rs
```diff
@@ -48,7 +48,7 @@ fn run_bench() {
 }
 
 async fn bench_task(tui: TUI, lines: Vec<String>) {
-    let msg_target = MsgTarget::Server { serv: "mentions" };
+    let msg_target = MsgTarget::Server { serv: "server" };
```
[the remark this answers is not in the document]
trevarj: It isn't I just did it for no reason other than "mentions" is a Misc tab now. I can revert it.

crates/libtiny_tui/src/tui.rs
trevarj: I thought you wanted this 😂

crates/tiny/src/ui.rs
trevarj: Agreed.

crates/libtiny_tui/src/tui.rs
```diff
@@ -626,6 +643,55 @@ impl TUI {
         tab.set_cursor(cursor);
     }
 
+    /// Shows "mentions" tab at the beginning of the tab list
+    pub(crate) fn show_mentions_tab(&mut self) -> Option<usize> {
```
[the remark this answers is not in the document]
trevarj: The issue is mainly because we can now /close the mentions tab and it can be recreated in `maybe_create_tab` when you are mentioned...Now that I think about it, it might just be better to not allow /close for the mentions tab. Thoughts?

crates/libtiny_tui/src/key_map.rs
```diff
@@ -152,6 +153,10 @@ impl KeyMap {
     pub(crate) fn load(&mut self, key_map: &KeyMap) {
         self.0.extend(key_map.0.iter())
     }
+
+    pub(crate) fn iter(&self) -> Iter<Key, KeyAction> {
```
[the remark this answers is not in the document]
trevarj: Thanks, I didn't think of that. Will keep it in mind when/if I'm creating public crates

crates/libtiny_tui/src/tui.rs
trevarj: I removed the automatic selection fully. Instead, I made the help tab highlight on creation.

osa1: Why? If I run `/help` I want to see help right now, without any more effort. Why run `/help` if you don't want to see it?

Why not switch to the tab in command implementation?

trevarj: Done.

crates/libtiny_tui/src/tui.rs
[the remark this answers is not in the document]
osa1: I just tried and I cannot close mentions tab in this branch. It seems like it's accidentally working?

I think it makes sense to not allow closing mentions tab because we rely on having at least one tab available in a few places, and changing all those places will require more design (for example, how to enter commands after closing all tabs) and implementation work.

crates/tiny/src/cmd.rs
```diff
@@ -159,6 +158,9 @@ fn close(args: CmdArgs) {
         MsgSource::User { serv, nick } => {
             ui.close_user_tab(&serv, &nick);
         }
+        // Don't allow closing the mentions tab
```
osa1, line 161: @trevarj you disallowed closing mentions tab here.

trevarj: Are you ok with being able to close it?

crates/tiny/src/cmd.rs
```diff
@@ -450,31 +464,36 @@ fn names(args: CmdArgs) {
...
+                        } else {
+                            ui.add_client_msg(&format!("{} is not in the channel", nick), &target);
+                        }
+                    }
+                } else {
+                    ui.add_client_err_msg(
+                        "/names only supported in chan tabs",
+                        &MsgTarget::CurrentTab,
+                    );
+                }
```
osa1, line 494: This diff should be much smaller if you implement this in the style of previous version:

```rust
let serv_name = match src.serv_name() {
    None => return,
    Some(serv_name) => serv_name,
};
```

Then the lines starting after `let client = ...` should not change at all.

crates/tiny/src/cmd.rs
osa1: No, see my previous comment.

crates/tiny/src/cmd.rs
```diff
@@ -518,16 +539,17 @@ static HELP_CMD: Cmd = Cmd {
...
                 cmd.name, cmd.description, cmd.usage
-            ),
-            &MsgTarget::CurrentTab,
-        )
-    }
+            )
+        })
+        .collect::<Vec<_>>();
+    ui.create_help_tab(&msgs);
+    ui.switch(MiscTab::Help.display());
```
osa1, line 552: I think this may break if we have a tab for a user with nick "help".

osa1:
I suggest changing this a little bit which will avoid problems like having a tab for a user with nick "help" and simplify the design quite a lot.

- Remove `close_misc_tab`. Make TUI handle `/close` if the tab is a "misc" tab. Otherwise the command should be returned to the caller as before.
- Add a `TUI::show_help` method. It should create a tab, add the help message, and switch to the tab. Important thing is to the caller it's now known how the help message is shown. It could be in the current tab, in an overlay, or in a new tab.
- In TUI when an `Input` event received check if the tab is "help". If so, show the error message as you do in this PR, and *ignore* the event.
- Once these are all implemented, I believe we no longer need the "misc tab" stuff, because tiny (the TUI event loop) will never receive an event with the help tab as the source.

I know I asked for too many changes already, sorry for that. I think this design will be much better. The key point is handling the help tab is entirely up to the TUI implementation, rest of the program does not need to know about it. We can also easily change how we show help in the future. With this I think only the files in libtiny_tui should change. The PR should not have any other changes.

Does this make sense?

Why not do the same for the mentions tab? I think maybe we could simplify some of the code by with a similar idea, but what's shown in mentions is decided by tiny (it searches every message for the current nick), so we it needs to know about the tab and manipulate it.

osa1:
Just to clarify my thought process here: showing a "help" dialogue (or a tab in our case) in a UI implementation should happen entirely in the UI implementation, should not effect elsewhere in the program. In our case, updating `MsgSource` is going against it. The plan I described above should make "help" tag an implementation detail of in the TUI and should not change interfaces (MsgSource, TUI API) in any way.

Looking at me previous message above

> Add a TUI::show_help method. It should create a tab, add the help message, and switch to the tab. Important thing is to the caller it's now known how the help message is shown. It could be in the current tab, in an overlay, or in a new tab.

I think we could handle `/help` in TUI and avoid this as well. We can pass the help string (or details needed to create a help string) to TUI on initialization and TUI can show it in `/help`.

trevarj:
Perhaps I'm being stubborn (or too tired to take another shot at this), but it might become messy since creating a new tab will require `MsgSource` to be optional. A lot of the functionality of `tui.rs` requires using a tab's `MsgSource`. Seems clunky to make `Tab` into an enum to support "help" tab functionality separately (no src but having a name/alias, having the ability to move freely between other tabs, etc), but also fragile to make "help" a server tab that has some exclusive behavior.

Maybe just keep current functionality (flood current tab with help), and simply the key mappings? 😬

osa1:
> it might become messy since creating a new tab will require MsgSource to be optional

There's only 4 uses of `TUI::new_tab`! `MsgSource` has a lot more uses (`ag --rust -w MsgSource` returns 84 occurrences) so changing `TUI::new_tab` should be much less messy than adding one more variant to `MsgSource`.

> A lot of the functionality of tui.rs requires using a tab's MsgSource

So the idea is most of these functions should just show the error message ("change tab to a server/chan/user") and return.

> Seems clunky to make Tab into an enum to support "help" tab functionality separately

We shouldn't need to turn `Tab` into an enum, that would be a large refactoring and I don't think it will buy us much. Probably make things worse.

> Maybe just keep current functionality (flood current tab with help), and simply the key mappings?

This would be fine too. I don't feel strongly either way.

To rephrase, `MsgSource` is a quite central type used everywhere. It doesn't make sense to add a variant to it for TUI-specific stuff. All we can do with that variant outside of TUI is to ignore it. Thinking about this more, I think it might make sense to move mentions tab stuff to TUI as well. TUI already knows the current nick so it can search in incoming messages for mentions and update the tab. No need for rest of the client to know about mentions tab. Same idea.

trevarj:
>  There's only 4 uses of TUI::new_tab

I meant that you actually can't create a `Tab` without a source, so when you change `src: Option<MsgSource>` a lot of matching on Some() will occur in other functions around tui.rs. Is that ok?

```
        self.tabs.insert(
            idx,
            Tab {
                alias,
                widget: MessagingUI::new(
                    self.width,
                    self.height - 1,
                    status,
                    self.scrollback,
                    self.msg_layout,
                ),
                src,
                style: TabStyle::Normal,
                switch,
                notifier,
            },
        );
```

trevarj:
WIP:
https://github.com/osa1/tiny/compare/master...trevarj:help-tab?expand=1

osa1:
Thanks @trevarj . If you could submit a PR it will be easier to find. You can create a PR as draft or mark it as draft after creating.

PR #341, opened 2021-07-30
Clarify that /reload is for TUI config, fix warnings

osa1:
@trevarj I can't add you as a reviewer but it would be great if you could take a look :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ Commands start with `/` character.
 - `/names`: List all nicks in the current channel. You can use `/names <nick>` to
   check if a specific nick is in the channel.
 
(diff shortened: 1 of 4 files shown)


trevarj (approved):
Looks good to me. Didn't realize `config_path` was never used here.

osa1:
> Didn't realize config_path was never used here.

Me too! I think it was used at some point, but then we moved TUI config stuff into the TUI crate and either something got lost in that refactoring (reloading server stuff), or we never really reloaded server stuff and this stuff became unused.

Since no one's complaining about `/reload` not reloading server config I'm guessing it's not used much so removing this dead code  should be OK.

PR #342, opened 2021-07-30
Fixed some clippy warnings

trevarj:
Others are false positives

### DIFF
--- a/crates/libtiny_client/src/state.rs
+++ b/crates/libtiny_client/src/state.rs
@@ -719,16 +719,14 @@ async fn retry_channel_join(
 
     let mut rcv_abort = ReceiverStream::new(rcv_abort).fuse();
 
-    match timeout(Duration::from_secs(10), rcv_abort.next()).await {
-        Err(_) => {
-            // Send join message
-            snd_irc_msg
-                .try_send(wire::join(std::iter::once(channel.as_ref())))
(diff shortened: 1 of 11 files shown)


crates/libtiny_client/src/state.rs
```diff
@@ -719,16 +719,14 @@ async fn retry_channel_join(
...
 
-    match timeout(Duration::from_secs(10), rcv_abort.next()).await {
-        Err(_) => {
-            // Send join message
-            snd_irc_msg
-                .try_send(wire::join(std::iter::once(channel.as_ref())))
-                .unwrap();
-        }
-        Ok(_) => {
-            // Channel tab was closed
```
osa1, line 730: Could you bring back this comment please? Maybe use put it before the `if`, something like

```
// Send join message after timeout. Error means channel tab was closed.
if timeout(...) { ... }
```

osa1 (approved):
Looks good, thanks @trevarj ! Left one minor comment.

crates/libtiny_client/src/state.rs
trevarj: Done. I think `Ok` means it was closed, right?

osa1: Right, thanks!

Issue #53, opened 2017-12-02
KICK message not handled


BrianOn99:
When a channel operator kick me out, tiny does not show it nor close the connection.  Interestingly, for some server, I can still send message to the channel.

osa1:
Thanks for reporting,

This is easy to implement once we figure out how to show it on the screen.

osa1:
I forgot that this issue existed and rediscovered it this weekend (thanks @trevarj)

If I kick another tiny user from a channel two things go wrong:

- The kicked user do not get notified in any way. As they see it, they're still in the channel and can post messages.
- Other tiny users in the channel do not get notified in any way, they still see that the kicked user is still in the channel.

Not sure if related but I also realized that when I have a privmsg tab to the kicked user and they join the channel again, I see a `+<nick>` line in the privmsg tab. I'd expect that line to be shown after the user left the server and then joined one of the channel we're in.

PR #344, opened 2021-08-02
Config to set /ignore and /notify

### LINKED ISSUE #118: how to setting /ignore and /notify by default?
subject

trevarj:
Added the ability to configure /ignore and /notify defaults for tabs.
The settings can be added under each chan, server or default section.

----

Overview:
- The most specific config will be the highest priority, i.e if you configure `ignore` under a chan then it will supersede the server's `ignore`, which will supersede the `ignore` under `defaults:`.

### Priority hierarchy (highest to lowest):
**Server tab:** server section -> defaults section -> hardcoded defaults (never ignore, notifier depends on conditional compilation, see `impl Default for Notifier`)
**Chan tabs:** chan section -> server section -> defaults section -> current setting on server tab (maybe someone toggled the setting manually)
**User tabs:** never ignore, notifier is always `Messages` (pre-existing behavior)

If acceptable, closes #118

### DIFF
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -4,6 +4,7 @@
 use libtiny_common::{ChanName, ChanNameRef};
 use serde::de::{self, Deserializer, MapAccess, Visitor};
 use serde::Deserialize;
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::Read;
 use std::path::Path;
@@ -35,55 +36,24 @@ pub(crate) struct Config {
     pub(crate) key_map: Option<KeyMap>,
 }
 
-impl Config {
-    /// Gets tab configs for `server`
-    /// Prioritizing configs under the server or using defaults
-    pub(crate) fn server_tab_configs(&self, server: &str) -> TabConfig {
-        let server_config = self
-            .servers
-            .iter()
-            .find(|s| s.addr == server)
-            .map(|s| &s.configs);
-        self.defaults.tab_configs.merge(server_config)
-    }
-
-    /// Gets tab configs for `chan` in `server`
-    /// Prioritizing configs under the chan, then the server, then the defaults
-    pub(crate) fn chan_tab_configs(&self, server: &str, chan: &ChanNameRef) -> TabConfig {
-        let tab_config = self
-            .servers
-            .iter()
-            .find(|s| s.addr == server)
-            .and_then(|s| s.join.iter().find(|c| chan == &c.name).map(|c| c.config));
-        self.server_tab_configs(server).merge(tab_config.as_ref())
-    }
-
-    pub(crate) fn user_tab_configs(&self) -> TabConfig {
-        TabConfig {
-            ignore: Some(false),
-            notifier: Some(Notifier::Messages),
-        }
-    }
-}
-
-#[derive(Debug, Deserialize, PartialEq)]
+#[derive(Debug, Deserialize, PartialEq, Eq)]
 pub(crate) struct Server {
     pub(crate) addr: String,
     pub(crate) join: Vec<Chan>,
     #[serde(flatten)]
-    pub(crate) configs: TabConfig,
+    pub(crate) config: Option<TabConfig>,
 }
 
-#[derive(Debug, Default, Deserialize, PartialEq)]
+#[derive(Debug, Default, Deserialize, PartialEq, Eq)]
 pub(crate) struct Defaults {
     #[serde(flatten)]
-    pub(crate) tab_configs: TabConfig,
+    pub(crate) tab_config: TabConfig,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Chan {
     pub name: ChanName,
-    pub config: TabConfig,
+    pub config: Option<TabConfig>,
 }
 
 impl<'de> Deserialize<'de> for Chan {
@@ -122,41 +92,80 @@ impl FromStr for Chan {
             s.to_string()
         };
         // Try to split chan name and args
-        match s.split_once(" ") {
+        match s.split_once(' ') {
             // with args
             Some((name, args)) => {
                 let configs = TabConfig::from_str(args)?;
                 Ok(Chan {
                     name: ChanName::new(name.to_string()),
-                    config: configs,
+                    config: Some(configs),
                 })
             }
             // chan name only
             None => Ok(Chan {
                 name: ChanName::new(s),
-                config: TabConfig::default(),
+                config: None,
             }),
         }
     }
 }
 
-#[derive(Debug, Default, Copy, Clone, Deserialize, PartialEq)]
+/// Map of TabConfigs by tab names
+#[derive(Debug, Default)]
+pub(crate) struct TabConfigs(HashMap<String, TabConfig>);
+
+impl TabConfigs {
+    pub(crate) fn serv_conf(&self, serv_name: &str) -> Option<TabConfig> {
+        self.0.get(serv_name).cloned()
+    }
+
+    pub(crate) fn chan_conf(&self, serv_name: &str, chan: &ChanNameRef) -> Option<&TabConfig> {
+        self.0.get(&format!("{}_{}", serv_name, chan.display()))
+    }
+}
+
+impl From<&Config> for TabConfigs {
+    fn from(config: &Config) -> Self {
+        let mut tab_configs = HashMap::new();
+        for server in &config.servers {
+            let serv_tc = config.defaults.tab_config.merge(server.config.as_ref());
+            tab_configs.insert(server.addr.clone(), serv_tc);
+            for chan in &server.join {
+                let tc = serv_tc.merge(chan.config.as_ref());
+                tab_configs.insert(format!("{}_{}", server.addr, chan.name.display()), tc);
+            }
+        }
+        Self(tab_configs)
+    }
+}
+
+#[derive(Debug, Default, Copy, Clone, Deserialize, PartialEq, Eq)]
 pub struct TabConfig {
     /// `true` if tab is ignoring join/part messages
     #[serde(default)]
-    pub ignore: Option<bool>,
+    pub ignore: bool,
     /// Notification setting for tab
     #[serde(default)]
-    pub notifier: Option<Notifier>,
+    pub notifier: Notifier,
 }
 
 impl TabConfig {
-    /// Overwrites `self`'s values with `o`'s if `o`'s are `Some`
-    pub(crate) fn merge(&self, o: Option<&TabConfig>) -> TabConfig {
-        if let Some(o) = o {
+    pub(crate) fn user_tab_config() -> TabConfig {
+        TabConfig {
+            ignore: false,
+            notifier: Notifier::Messages,
+        }
+    }
+
+    pub(crate) fn merge(&self, config: Option<&TabConfig>) -> TabConfig {
+        if let Some(config) = config {
             TabConfig {
-                ignore: o.ignore.or(self.ignore),
-                notifier: o.notifier.or(self.notifier),
+                ignore: self.ignore ^ config.ignore,
+                notifier: if config.notifier > self.notifier {
+                    config.notifier
+                } else {
+                    self.notifier
+                },
             }
         } else {
             self.to_owned()
@@ -174,13 +183,13 @@ impl FromStr for TabConfig {
             .try_fold(TabConfig::default(), |mut tc, arg| match arg {
                 // flag
                 "ignore" => {
-                    tc.ignore = Some(true);
+                    tc.ignore = true;
                     Ok(tc)
                 }
                 arg => match arg.split_once(' ') {
                     // arg with parameter
                     Some(("notify", val)) => {
-                        tc.notifier = Some(Notifier::from_str(val)?);
+                        tc.notifier = Notifier::from_str(val)?;
                         Ok(tc)
                     }
                     _ => Err(format!("Unexpected argument: {:?}", arg)),
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -39,13 +39,11 @@ impl Config {
     /// Gets tab configs for `server`
     /// Prioritizing configs under the server or using defaults
     pub(crate) fn server_tab_configs(&self, server: &str) -> TabConfig {
-        let server_config = self.servers.iter().find_map(|s| {
-            if s.addr == server {
-                Some(&s.configs)
-            } else {
-                None
-            }
-        });
+        let server_config = self
+            .servers
+            .iter()
+            .find(|s| s.addr == server)
+            .map(|s| &s.configs);
         self.defaults.tab_configs.merge(server_config)
     }
 
@@ -56,15 +54,7 @@ impl Config {
             .servers
             .iter()
             .find(|s| s.addr == server)
-            .and_then(|s| {
-                s.join.iter().find_map(|c| {
-                    if &c.name == chan {
-                        Some(c.config)
-                    } else {
-                        None
-                    }
-                })
-            });
+            .and_then(|s| s.join.iter().find(|c| chan == &c.name).map(|c| c.config));
         self.server_tab_configs(server).merge(tab_config.as_ref())
     }
 
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -220,7 +220,7 @@ pub(crate) enum Layout {
     Aligned,
 }
 
-#[derive(Clone, Debug, Deserialize)]
+#[derive(Debug, Deserialize)]
 #[serde(default)]
 pub struct Colors {
     pub nick: Vec<u8>,
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -162,6 +162,16 @@ impl TabConfigs {
         };
         self.0.insert(key, config);
     }
+
+    pub(crate) fn set_by_server(&mut self, serv_name: &str, config: TabConfig) {
+        for c in self
+            .0
+            .iter_mut()
+            .filter(|entry| entry.0.starts_with(serv_name))
+        {
+            *c.1 = config;
+        }
+    }
 }
 
 impl From<&Config> for TabConfigs {
@@ -207,6 +217,12 @@ impl TabConfig {
             ignore: self.ignore.or(config.ignore),
             notify: self.notify.or(config.notify),
         }
+    }
+
+    pub(crate) fn toggle_ignore(&mut self) -> bool {
+        let ignore = self.ignore.get_or_insert(false);
+        *ignore = !&*ignore;
+        *ignore
     }
 }
 
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -9,7 +9,7 @@ use std::io::Read;
 use std::path::Path;
 use std::str::FromStr;
 
-pub(crate) use termbox_simple::*;
+pub use termbox_simple::*;
 
 use crate::key_map::KeyMap;
 use crate::notifier::Notifier;
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -1,6 +1,7 @@
 // To see how color numbers map to actual colors in your terminal run
 // `cargo run --example colors`. Use tab to swap fg/bg colors.
 
+use libtiny_common::ChanNameRef;
 use serde::de::{self, Deserializer, MapAccess, Visitor};
 use serde::Deserialize;
 use std::fs::File;
@@ -10,9 +11,14 @@ use std::path::Path;
 pub use termbox_simple::*;
 
 use crate::key_map::KeyMap;
+use crate::notifier::Notifier;
 
-#[derive(Deserialize)]
+#[derive(Debug, Default, Deserialize)]
 pub(crate) struct Config {
+    pub(crate) servers: Vec<Server>,
+
+    pub(crate) defaults: Defaults,
+
     #[serde(default)]
     pub(crate) colors: Colors,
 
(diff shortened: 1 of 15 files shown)


crates/tiny/src/config.rs
```diff
@@ -41,7 +43,7 @@ pub(crate) struct Server {
 
     /// Channels to automatically join.
     #[serde(default)]
-    pub(crate) join: Vec<String>,
+    pub(crate) join: Vec<Chan>,
```
trevarj, line 46: Can replace with `libtiny_tui`'s `Chan`, if we want to export that type, or keep it like this.

crates/libtiny_tui/src/config.rs
```diff
@@ -41,14 +158,14 @@ pub struct Style {
...
-#[derive(Deserialize)]
+#[derive(Debug, Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub(crate) enum Layout {
     Compact,
     Aligned,
 }
 
-#[derive(Debug, Deserialize)]
+#[derive(Clone, Debug, Deserialize)]
```
osa1, line 168: Unused `Clone`

crates/libtiny_tui/src/config.rs
```diff
@@ -28,6 +34,117 @@ pub(crate) struct Config {
...
+    /// Channel specified by name and extra configurations
+    WithConfigs {
+        name: String,
+        #[serde(flatten)]
+        configs: TabConfigs,
+    },
+}
+
+#[derive(Debug, Default, Copy, Clone, Deserialize, PartialEq)]
+pub(crate) struct TabConfigs {
```
osa1, line 109: Maybe make this singular, `TabConfig`? Maybe that's just me but I read this as "configurations for multiple tabs", but it's for one tab.

crates/libtiny_tui/src/config.rs
```diff
@@ -28,6 +34,117 @@ pub(crate) struct Config {
...
+    /// `true` if tab is ignored
```
osa1, line 110: Would be good to say more about what "ignore" means in this context. Maybe add "(ignore = don't show joins and parts/leaves)". Or maybe just say that this is for `/ignore`, then the reader can refer to the cmd documentation.

osa1 (commented):
Thanks @trevarj . Just did a quick pass, it looks great.

> No support for /join or /connect command args

I wonder if we should also do that. We will need to implement command args anyway, and if we do that today the fields `ignore` and `notify` will be redundant, so perhaps we could just rely on command args, instead of having two ways of doing the same thing in the config (which raises questions like which one will have priority when in conflict).

If we can design a good syntax for command args then it should be a few lines of change in the config parser. We could use the same types and rest of the code.

trevarj:
@osa1 my problem with the command args approach was that it's not straightforward going from `cmd.rs` all the way to the UI with the current way we handle auto-joining channels. I believe we would have to keep UI specific configs in `State`, which seemed slightly hacky to me at the time. Maybe you know a better way, or if that is acceptable?

crates/tiny/src/cmd.rs
```diff
@@ -280,7 +280,7 @@ static JOIN_CMD: Cmd = Cmd {
     name: "join",
     cmd_fn: join,
     description: "Joins a channel",
-    usage: "`/join <chan1>,<chan2>,...` or `/join` in a channel tab to rejoin",
+    usage: "`/join <chan1> <chan2>,...` or `/join` in a channel tab to rejoin",
```
osa1, line 283: The last comma should be gone too

crates/tiny/src/config.rs
```diff
@@ -64,6 +66,28 @@ pub(crate) struct Defaults {
...
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub(crate) enum Chan {
+    /// Channel specified by name only
+    Name(String),
+    /// Channel specified by name with extra configuration (used by tui crate)
+    WithExtra {
+        name: String,
+        #[serde(flatten)]
+        extra: HashMap<String, Value>,
```
osa1, line 78: Why is this needed? I think TUI is parsing the config file again, it's not using this, no?

trevarj: To make the `tiny` `Config` generic and the `libtiny_tui` `Config` aware of the specific values.
My comment above:
> Can replace with libtiny_tui's `Chan` (requires exporting the type), if we want to export that type, or keep it like this.

trevarj:
I was able to add parsing for `/join` and switched the config's `join:` to an identical string, i.e `- "#tiny -ignore"`, but ran into an issue...
After adding a new UI function, `set_tab_config(..)`, which I wanted to call from in `cmd.rs` after `client.join()`, I realized that there's an annoying race.
1. `/join #tiny -ignore` is entered
2. parse the command args, call `client.join()` as normal
3. call `ui.set_tab_config()` to set the proper ignore value https://github.com/trevarj/tiny/blob/default-ignore-notify-v2/crates/tiny/src/cmd.rs#L339
Turns out `ui.set_tab_config()` runs through, then gets clobbered by the client joining and creating a new tab. Doesn't matter if I switch the ordering or set `set_tab_config()` to `can_create_tab`

Need to think about this more I guess. Branch is here: https://github.com/trevarj/tiny/tree/default-ignore-notify-v2

trevarj:
From most recent commits: 

> When joining a channel using the /join command, the tab is created
before the server is actually joined so we can effectively set the
channel config (ex. ignore the tab, set notifier).
Due to this, I had to add handling for REPLY messages that occur when a
channel cannot be joined but the tab has already been created. Now an
error message shows on the chan tab, giving the reason why the chan
can't be joined.

@osa1 let me know if you get some time to try this out and can give some thoughts on the UX here.

crates/libtiny_tui/src/config.rs
```diff
@@ -1,18 +1,25 @@
...
+use libtiny_common::{ChanName, ChanNameRef};
 use serde::de::{self, Deserializer, MapAccess, Visitor};
 use serde::Deserialize;
 use std::fs::File;
 use std::io::Read;
 use std::path::Path;
+use std::str::FromStr;
 
-pub use termbox_simple::*;
+pub(crate) use termbox_simple::*;
```
osa1, line 12: Huh, I wonder how this slipped through.. It makes no sense to export termbox stuff from the `config` module. I think we should be able to remove the `pub` entirely, but maybe not in this PR. In fact, reverting this might be good as it doesn't seem to be relevant to the feature.

crates/libtiny_tui/src/tui.rs
```diff
@@ -377,6 +384,38 @@ impl TUI {
             ret
         };
 
+        // Get tab configs for the type of tab being created
+        let TabConfig { ignore, notifier } = if let Some(config) = self.load_config() {
```
osa1, line 388: This will read and parse the config file every time we crate a tab, right? Why not just store the config somewhere and only reload when asked with `/reload`?

crates/libtiny_tui/src/tui.rs
```diff
@@ -377,6 +384,38 @@ impl TUI {
...
+                        .find_map(|tab| {
+                            if tab.src.serv_name() == serv {
+                                Some(TabConfig {
+                                    ignore: Some(!tab.widget.is_showing_status()),
+                                    notifier: Some(tab.notifier),
+                                })
+                            } else {
+                                None
+                            }
+                        })
```
osa1, line 406:
```suggestion
                    let server_tab_config = self
                        .tabs
                        .iter()
                        .find(|tab| tab.src.serv_name() == serv)
                        .map(|tab| TabConfig {
                            ignore: Some(!tab.widget.is_showing_status()),
                            notifier: Some(tab.notifier),
                        })
```
 
Same for other uses of `find_map`.

crates/libtiny_tui/src/config.rs
```diff
@@ -28,6 +35,171 @@ pub(crate) struct Config {
     pub(crate) key_map: Option<KeyMap>,
 }
 
+impl Config {
+    /// Gets tab configs for `server`
+    /// Prioritizing configs under the server or using defaults
+    pub(crate) fn server_tab_configs(&self, server: &str) -> TabConfig {
+        let server_config = self.servers.iter().find_map(|s| {
```
osa1, line 42: `find_map` here can be simplified as shown in my other comment.

crates/libtiny_tui/src/config.rs
```diff
@@ -28,6 +35,171 @@ pub(crate) struct Config {
...
+
+    /// Gets tab configs for `chan` in `server`
+    /// Prioritizing configs under the chan, then the server, then the defaults
+    pub(crate) fn chan_tab_configs(&self, server: &str, chan: &ChanNameRef) -> TabConfig {
+        let tab_config = self
+            .servers
+            .iter()
+            .find(|s| s.addr == server)
+            .and_then(|s| {
+                s.join.iter().find_map(|c| {
```
osa1, line 60: Ditto.

crates/tiny/config.yml
```diff
@@ -29,13 +34,25 @@ servers:
       # (useful when `pass` or `sasl` fields above are not used)
       # nickserv_ident: 'hunter2'
 
+      # Set /ignore for this server and all its tabs, default: false
+      # ignore: true  
+
+      # Set notifier for this server and all its tabs, default: mentions
+      # notifier: <off | mentions | messages>
```
osa1, line 41: Syntax here is inconsistent with the `join` example above. Above we use `[off | mentions | messages]`, here `<off | mentions | messages>`. I don't have an opinion on which one makes more sense but we should pick one and be consistent. I wonder which one of these syntax is more common?

crates/tiny/config.yml
```diff
@@ -29,13 +34,25 @@ servers:
...
     nicks: [tiny_user]
     realname: yourname
     join: []
     tls: false
 
+    # Set /ignore for all servers and all its tabs, default: false
+    # ignore: true 
+
+    # Set notifier for all servers and all its tabs, default: mentions
+    # notifier: <off | mentions | messages>
```
osa1, line 54: We use `<...>` here again.

crates/tiny/src/cmd.rs
```diff
@@ -280,7 +282,7 @@ static JOIN_CMD: Cmd = Cmd {
     name: "join",
     cmd_fn: join,
     description: "Joins a channel",
-    usage: "`/join <chan1>,<chan2>,...` or `/join` in a channel tab to rejoin",
+    usage: "`/join <chan1> [-ignore] [-notify [off|mentions|messages]],<chan2>...` or `/join` in a channel tab to rejoin",
```
osa1, line 285: Here we use `[...]` syntax..

crates/tiny/src/conn.rs
```diff
@@ -478,91 +478,117 @@ fn handle_irc_msg(ui: &UI, client: &dyn Client, msg: wire::Msg) {
...
+                                false,
+                            );
+                            ui.set_tab_style(TabStyle::NewMsg, &msg_target);
+                        }
+                        Some(User { .. }) | None => {
+                            debug!(
+                                "Ignoring numeric reply {}: pfx={:?}, params={:?}",
+                                n, pfx, params
+                            );
+                        }
```
osa1, line 591: I feel like changes in this file are unrelated to this PR, right? Could you revert these please? I'm happy to review these changes in its own PR.

osa1 (commented):
Thanks @trevarj . Just one quick pass for now, will review again in detail later. It would be good to revert unrelated changes in the meantime.

One question re: PR description:

>  Chan tabs: chan section -> server section -> defaults section -> current setting on server tab (maybe someone toggled the setting manually)

Shouldn't current tab setting have highest priority?

crates/tiny/config.yml
trevarj: I have seen `< blah | blah | blah >` used in `StructOpt`, so maybe that?

crates/libtiny_tui/src/tui.rs
trevarj: Yeah, that does make more sense. Should I add that in this PR?

osa1: I'd prefer doing it in this PR, yes. Otherwise master will have this issue for a while. It will be best if we never introduce it in the first place.

trevarj:
> Shouldn't current tab setting have highest priority?

@osa1 Once I store the configs (re: your comment about not loading every time) I will toggle the stored `TabConfig` when the user manually changes one of the parameters (with /ignore or /notify). 

I will update the PR description on my next commit

crates/libtiny_tui/src/config.rs
```diff
@@ -28,6 +35,161 @@ pub(crate) struct Config {
...
+impl FromStr for Chan {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // Make sure channel starts with '#'
+        let s = if !s.starts_with('#') {
+            format!("#{}", s)
+        } else {
+            s.to_string()
+        };
```
osa1, line 123: Not a big deal probably, but I wonder if we should expect user to get this right instead of inserting a `#` for them. The problem is channel names can contain multiple `#`s, not just one, but we assume one here.

crates/libtiny_tui/src/config.rs
```diff
@@ -28,6 +35,161 @@ pub(crate) struct Config {
...
+    #[serde(default)]
+    pub ignore: Option<bool>,
+    /// Notification setting for tab
+    #[serde(default)]
+    pub notifier: Option<Notifier>,
+}
+
+impl TabConfig {
+    /// Overwrites `self`'s values with `o`'s if `o`'s are `Some`
+    pub(crate) fn merge(&self, o: Option<&TabConfig>) -> TabConfig {
```
osa1, line 155: What does 'o' stand for here?

crates/tiny/src/conn.rs
```diff
@@ -544,6 +544,37 @@ fn handle_irc_msg(ui: &UI, client: &dyn Client, msg: wire::Msg) {
...
+            {
+                let chan = &params[1];
+                let msg = &params[2];
+                ui.add_client_err_msg(
+                    msg,
+                    &MsgTarget::Chan {
+                        chan: ChanNameRef::new(chan),
+                        serv,
+                    },
+                );
```
osa1, line 577: Revert unrelated changes please. Didn't you already revert some of this file? Why keep this?

Happy to review these changes in their own PR.

osa1 (commented):
We should document how overriding `notify` and `ignore` (e.g. tab `ignore` overrides `server` and `default`). I'll post a suggestion here.

crates/libtiny_tui/src/config.rs
trevarj: This is something I noticed other clients do, so I added it if there is no `#`. It seems like a nicer UX, but we don't need to have it.

crates/tiny/src/conn.rs
trevarj: These are actually needed for this PR.  Since it is necessary to create the tab from`/join`(in order to set the tab configs) BEFORE the actual JOIN message is sent over the wire, we have to support these for when you can't join the channel. It also adds a nicer UX, in my opinion, when you can't join a channel that you added to your config file. It will spawn the tab but tell you that the reason why you couldn't join.

I'm a little lost in this whole feature's design/implementation now, since I could probably get around this issue by storing the tab config on the tui...can be like this:

Right now it's like:
1. `/join #blah -ignore`
2. spawns #blah tab and sets ignored
3. JOIN sent over the wire 
4. Joined or might be an error so show it on the tab

Can be like this instead:
1. /join #blah -ignore
2. triggers `ui.set_tab_config()`...stores a TabConfig for blah channel on the TUI (probably will be storing configs anyway)
3. JOIN sent over the wire
4. #blah tab is spawned since that's what happens anyway

osa1: OK, I see the motivation now. I still think these should be in a separate PR though. Reasons:

- These messages should be handled even without this PR. How do we show these errors currently? I hope we're not just ignoring them.
- This PR would work without handling these messages. UX would not be ideal, but I think it also wouldn't be any worse than before. Only case where without these changes this PR would make things worse is if we currently crate a tab and show these errors there, but I doubht that is the case.

Do you know what we do for these messages currently, @trevarj?

trevarj: They get logged but just ignored. I am working on cleaning this up.

osa1: > They get logged but just ignored. I am working on cleaning this up.

OK, this is bad. Thanks for working on this.

Btw, another similar issue is #53. I don't remember how bans work on IRC but if there are ban messages we should also support them while we're at it.

Raphalex46:
Hi, is this PR still going to be merged ? I was looking to know if it was possible to set `notify` in the `config.yml` and arrived here. If you need help with anything I can take a look. Being able to set default configuration values in the configuration file could be a very useful feature.

trevarj:
@Raphalex46 I've rebased with `master` to fix the conflicts. If you'd like you can test this out and see if it works as you expect. I have been using it for a long time now and it works well for me. See [here](https://github.com/trevarj/tiny/blob/default-ignore-notify/crates/tiny/config.yml#L17) for config details

Raphalex46:
I've tested it a little bit, the only that I didn't manage to get working is the `-notify off` option.

trevarj:
@Raphalex46 I made another commit with some fixes. You can do `/reload` to reload the configs. 
Make sure you compiled with `features = desktop-notifications`.

Raphalex46:
I can't test it right now but I will as soon as I can and I'll let you know then.
Yes it was compiled with `features=desktop-notifications`, I just didn't manage to get the
notifications to "off" by default.

trevarj:
>  I just didn't manage to get the notifications to "off" by default

Right, should work now.

Raphalex46:
Yup, seems to work pretty much as expected. I suppose we have to wait to see if @osa1 is ok with merging this.

I was also thinking, eventually, options like these could be expressed in the config file like this:
```yml
join:
  - name: "#tiny"
    ignore: true
    notify: messages
```
right ?

That way, configuration defaults could even be expressed for a whole server:
```yml
servers:
  - addr: irc.oftc.net
    port: 6697
    # ...
    ignore: true
    notify: messages
```

trevarj:
@Raphalex46 thanks a lot for taking the time to test.

> I was also thinking, eventually, options like these could be expressed in the config file like this:

The reason why they are using a special parsing was so that it could be leveraged for on-the-fly `/connect` or `/join`, e.g if you want to join a chan and have it ignored in one shot (kinda pointless, but saves a click...). That actually works now, if you do `/help` you should see it in the help message for `/join`.


> That way, configuration defaults could even be expressed for a whole server:

This should actually be working too.
From [this test](https://github.com/trevarj/tiny/blob/default-ignore-notify/crates/libtiny_tui/src/tests/config.rs#L8-L21) in the code you can see a simple example. The precedence is chan -> server -> defaults. 

- If you create a window with `/connect` it should use the configs under the server, or fallback to defaults.
- If you create a window with `/join` it should use the config of the chan, fallback to the current server tab config (may be modified on the fly by user), and then defaults.

I will write more thorough docs in the example `config.yaml` once osa1 gives the 👌🏻.
Bit lazy to do it right now especially since this PR was sitting for over a year!!

Raphalex46:
Yes, I saw that you can join like that, and it totally justifies the parsing!

I also did not see that you could put defaults like that for servers in the config file, so thanks for pointing that out.

I just mentioned this because we have a YAML config file, and it seemed more satisfying to use that for the default config of the channels. You're right that this PR probably isn't the place to implement this, just wanted to put it out there before I forgot :)

crates/tiny/src/cmd.rs
```diff
@@ -313,16 +317,30 @@ fn join(args: CmdArgs) {
...
+                None
+            }
+        })
+        .collect::<Vec<Chan>>();
 
-    let chans: Vec<ChanName> = if words.is_empty() {
+    let chans = if chans.is_empty() {
         match ui.current_tab() {
             None => return,
+            // rejoin current tab channel
```
osa1, line 338: "tab channel" sounds a bit strange to me. Maybe "rejoin current channel" or "rejoin current tab's channel"?

crates/tiny/src/conn.rs
osa1: @trevarj Could you move this change to its own PR please? It's useful on its own, we should just merge it. That will also make reviewing the PR easier.

crates/libtiny_tui/src/tests/config.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::config::*;
+use crate::notifier::Notifier;
+
+#[test]
+fn parsing_tab_configs() {
+    let config_str = r##"
+        servers:
+          - addr: "server"
+            join: 
+              - "#tiny -ignore -notify messages"
```
osa1, line 12: I don't like this syntax in config, it's inconsistent with how we set the same things for servers and globally.

We should use the same `notifier` and `ignore` fields for all.

Maybe something like:

```
join:
  - name: "#tiny"
    ignore: false
    notifier: off
```

The current syntax should also be supported.

We should also use same naming in the config files and commands. So "notify" instead of "notifier".

osa1 (commented):
Thanks again @trevarj!

Added some inline comments.

There are lots of formatting changes in the PR, could you revert those please?

I know we're inconsistent in ordering of imports, but for now if you could follow the conventions in files to keep the diff small that would be really appreaciated. We can then do a pass to format them if we want.

(I'll add a few more things)

crates/libtiny_tui/src/tests/config.rs
trevarj: Yes, fair enough. I will do this.

osa1 (commented):
As discussed in #397:

- We should store all tab settings in TUI settings so that when we need to create a tab we will be able to find the settings for the tab.

  We still need to set tab settings on creation, to be able to implement `/join ... -ignore` etc. commands.
- **Design question:** if I update settings of a tab, then close it, and create the tab again, should I get the settings in the config or the updated settings?
  - **Option (1) Use the updated settings:** We should not store the tab settings in `Tab`, get the settings from the config every time we need it.
  - **Option (2) Use the settings from the config file:** We should store the tab settings in `Tab`
  - Currently notification settings are stored in `Tab`, but for `ignore` we use the config. This needs to be consistent: either use both from the config (implements (1)), or store both in `Tab` (implements (2)).

I don't feel strongly, but I think (1) is probably better.

trevarj:
> Design question: if I update settings of a tab, then close it, and create the tab again, should I get the settings in the config or the updated settings?

Let's just go with option (1) so that the previous config for the tab is restored from `tab_configs` on `TUI` (which was sourced from the config file at startup). This seems like an unusual edge case anyway.

I will make an attempt at the following:
- [ ] Move tab config state (`notifier`, `status`/`ignore`) out of `Tab` and only have `TabConfigs` on `TUI`
- [ ] Remove all formatting changes
- [ ] Use proper yaml for chan config under `join:` (if we change `join` it will break everyone's config)

trevarj:
The above checklist is done, albeit with a bit of a hack to resolve an issue that I had that comes from moving `ignore` and `notifier` off of `Tab` (has to do with toggling if you look into the code).

This has lost its fun for me, so if anyone wants to help, now is the time 😩

Raphalex46:
I will try to take a look when I have the time, probably this week-end.

osa1:
Thanks @trevarj, I think this looks pretty good. Let's test this a little bit (the parts without unit tests) and merge. I will update README and CHANGELOG.

mizlan:
Is there anything blocking merge?

crates/libtiny_tui/src/tests/config.rs
osa1: It looks like this was done. Resolving.

osa1:
Sorry for the delay. I'll try to merge this this week.

PR #351, opened 2021-09-05
Add recalculating scroll on window resize

### LINKED ISSUE #319: Invalid scroll value in compact layout after resizing

trevarj:
Closed previous PR (#320) since I found an issue with it while playing on my experimental branch. This solution is much simpler and behaves the way that irssi does. 

- Recalculate scroll value based on new total visible lines after a resize
- Moved tests and added new ones

Fixes #319

### DIFF
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -139,7 +139,7 @@ impl MsgArea {
             Some(height) => height,
             None => {
                 let mut total_height = 0;
-                for line in self.lines.iter_mut() {
+                for line in &mut self.lines {
                     total_height += line.rendered_height(self.width);
                 }
                 self.lines_height = Some(total_height);
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -48,12 +48,12 @@ impl Layout {
 
 #[derive(Debug, Default)]
 struct Scroll {
-    /// An offset from the last visible line.
-    /// E.g. when this is 0, `self.lines[self.lines.len() - 1]` is drawn at the
-    /// bottom of screen.
+    /// An offset of visible lines from the last visible line of the most recent Line
+    /// E.g. when this is 0, `self.lines.last()` is drawn at the
+    /// bottom of screen. When it is not 0, it is N lines up in the message area.
     scroll: i32,
-    /// Current index of line drawn at the top of the screen
-    /// Used to recalculate `scroll` on resizing of window
+    /// Current index into `self.lines` of the Line drawn at the top of the screen.
+    /// Used to recalculate `scroll` on resizing of window and maintain scrolling in the correct place
     line_idx: usize,
 }
 
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -17,8 +17,10 @@ pub(crate) struct MsgArea {
     width: i32,
     height: i32,
 
-    /// Vertical scroll
-    scroll: Scroll,
+    /// Vertical scroll: An offset from the last visible line.
+    /// E.g. when this is 0, `self.lines[self.lines.len() - 1]` is drawn at the
+    /// bottom of screen.
+    scroll: i32,
 
     line_buf: Line,
 
@@ -46,17 +48,6 @@ impl Layout {
     }
 }
 
-#[derive(Debug, Default)]
-struct Scroll {
-    /// An offset of visible lines from the last visible line of the most recent Line
-    /// E.g. when this is 0, `self.lines.last()` is drawn at the
-    /// bottom of screen. When it is not 0, it is N lines up in the message area.
-    scroll: i32,
-    /// Current index into `self.lines` of the Line drawn at the top of the screen.
-    /// Used to recalculate `scroll` on resizing of window and maintain scrolling in the correct place
-    line_idx: usize,
-}
-
 impl MsgArea {
     pub(crate) fn new(width: i32, height: i32, scrollback: usize, layout: Layout) -> MsgArea {
         MsgArea {
@@ -64,7 +55,7 @@ impl MsgArea {
             scrollback,
             width,
             height,
-            scroll: Scroll::default(),
+            scroll: 0,
             line_buf: Line::new(),
             lines_height: Some(0),
             layout,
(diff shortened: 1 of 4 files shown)


crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -67,9 +67,14 @@ impl MsgArea {
     }
 
     pub(crate) fn resize(&mut self, width: i32, height: i32) {
-        self.width = width;
+        self.width = if self.scroll != 0 { width - 1 } else { width };
```
osa1, line 70: How does this work? Why do you make the area smaller in width if not scrolled to the bottom?

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -67,9 +67,14 @@ impl MsgArea {
...
+        let old_height = self.height;
         self.height = height;
+        let old_total_lines = self.lines_height;
```
osa1, line 73: Not in this PR but we should probably rename `lines_height` to `total_lines` or something else that describes it better. The fact that you used `old_total_lines` and not `old_lines_height` suggests `line_height` is not a good name for it.

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -67,9 +67,14 @@ impl MsgArea {
...
         self.lines_height = None;
+
+        self.total_visible_lines();
```
osa1, line 76: This method name looks like it's going to return something but from this use it seems like it's doing an update? Maybe the name should reflect that, e.g. `update_total_visible_lines` (no idea what the method is doing yet, just describing the problem).

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -128,12 +133,13 @@ impl MsgArea {
...
 impl MsgArea {
-    fn lines_height(&mut self) -> i32 {
+    /// The total number of visible lines if each Line was rendered at the current screen width
+    fn total_visible_lines(&mut self) -> i32 {
         match self.lines_height {
             Some(height) => height,
             None => {
                 let mut total_height = 0;
-                for line in &mut self.lines {
+                for line in self.lines.iter_mut() {
```
osa1, line 142: Why change this?

I'm a bit surprised that we update all lines (or "messages" as I suggest above) here. If I'm currently showing lines/messages  in range `n..m` (in Rust range syntax), and make the messaging area larger, most of the time I only need to know the a few lines before `n` and/or after `m` (depending on the implementation). Similarly when making it smaller, I'll need to update a smaller range in `n..m` (e.g. `n+x..m-y` where `x, y >= 0`).

So it's wasteful to update all the lines here. I think most of the values computed won't even be used most of the time.

Not really related to this PR though, I will open an issue about this.

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -128,12 +133,13 @@ impl MsgArea {
 // Scrolling
 
 impl MsgArea {
-    fn lines_height(&mut self) -> i32 {
+    /// The total number of visible lines if each Line was rendered at the current screen width
+    fn total_visible_lines(&mut self) -> i32 {
```
osa1, line 137: I found this method a bit confusing. I think we need different naming for the messages added to the messaging area and lines shown in TUI. Currently I think we call both "lines". Perhaps we could call the former "message" and the latter "line". Messages do not change as we resize, but lines do. Then name of this functions becomes less confusing. It would look like this:

```rust
/// Total number of visible lines when each `Message` rendered at the current messaging area width
fn total_visible_lines(...) { ... }
```

osa1:
Sigh.. GitHub lost my top-level comment when I went back to adding inline comments and then submitted.

> This solution is much simpler and behaves the way that irssi does.

For those of us not familiar with irssi, could you describe the idea? If irssi has any documentation quoting it or even linking to it (with a persistent link) would also be good.

crates/libtiny_tui/src/msg_area/mod.rs
trevarj: This was a merge error. I have a lot of crazy branches going and trying to sync them! I will remove this.

FYI it was for making room for a scrollbar

crates/libtiny_tui/src/msg_area/mod.rs
trevarj: Yes, there can definitely be a more efficient way to handle resizing/scrolling.

trevarj:
@osa1 well, after I tested more I don't think we are quite doing what irssi does. When scrolling, they seem to have a way of locking onto the top-most message that is visible. So, once you resize you are locked to that place. I think it's really bad that the screen goes black though, and this does fix that!

What I am doing is maintaining the scroll ratio (i.e if I scroll half way, resizing will keep me at half way, but not the location I was at) 

I tried to capture the differences, but it's difficult to see I think.
On both programs, I scrolled up a little and resized.
irssi-vs-tiny

osa1:
@trevarj could you move renaming to another PR please? It's difficult to see changes in resizing now with all the other changes. ("Not in this PR but we should probably rename ...")

osa1:
Screenshot_2021-09-06_19-21-26


Not sure how to reproduce yet but if I scroll up a little bit (maybe just one message) then randomly resize the window I get an empty line at the bottom (the highlighted line in the screenshot) even though there are lines to show at the top.

osa1:
Here's another screenshot with the bug:

Screenshot_2021-09-07_18-37-02

osa1:
@trevarj here's how to reproduce: make the window small in height so that the text will be scrolled. Move one line up with shift+up. Then make the window larger 5 or more rows. Do not make it one row larger at a time, it works fine if you add one row. You need to add a few rows. How to do that depends on the window manager, I guess. I'm using i3, so in resize mode if I make it larger it expands a few rows. I will record a gif if I can figure out how.

osa1 (approved):
Looks great, thanks again @trevarj !

Tests are using compact layout. Do you think the layout does not matter for these tests or should we maybe make them use aligned layout, or maybe have two variants for each test? We could refactor the test code a little bit to pass two expected screen arguments to `expect_screen`, one for the compact layout, one for aligned.

trevarj:
> We could refactor the test code a little bit to pass two expected screen arguments to expect_screen, one for the compact layout, one for aligned.

Can we do that in a separate issue?

PR #356, opened 2021-09-13
TUI: Fix line heights on `/resize`

### LINKED ISSUE #355: Panic after `/clear`
I run `/clear` today on a tab and after a few minutes got this panic:

Screenshot_2021-09-13_08-18-34

Unfortunately I didn't have backtraces enabled but I'll enable it now and try again. I'll update if it panics again.

trevarj:
When you /clear the screen the total lines in the MsgArea gets cleared
and set to None. Then if you resize after that, we were unwrapping total
lines to recalculate the scroll amount and panicking.

We know the total line height is 0 after /clear so we now set total line
heights `Some(0)`.

Fixes #355

### DIFF
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -74,9 +74,7 @@ impl MsgArea {
         self.lines_height = None;
 
         self.update_total_visible_lines();
-        if let Some(old_total_lines) = old_total_lines {
-            self.recalculate_scroll(old_height, old_total_lines);
-        }
+        self.recalculate_scroll(old_height, old_total_lines.unwrap());
     }
 
     pub(crate) fn layout(&self) -> Layout {
--- a/crates/libtiny_tui/src/msg_area/mod.rs
+++ b/crates/libtiny_tui/src/msg_area/mod.rs
@@ -74,7 +74,9 @@ impl MsgArea {
         self.lines_height = None;
 
         self.update_total_visible_lines();
-        self.recalculate_scroll(old_height, old_total_lines.unwrap());
+        if let Some(old_total_lines) = old_total_lines {
+            self.recalculate_scroll(old_height, old_total_lines);
+        }
     }
 
     pub(crate) fn layout(&self) -> Layout {
@@ -228,7 +230,7 @@ impl MsgArea {
     pub(crate) fn clear(&mut self) {
         self.lines.clear();
         self.scroll = 0;
-        self.lines_height = None;
+        self.lines_height = Some(0);
     }
 }
 
--- a/crates/libtiny_tui/src/tests/resize.rs
+++ b/crates/libtiny_tui/src/tests/resize.rs
@@ -315,3 +315,45 @@ fn test_resize_scroll_resize() {
 
     expect_screen(screen1, &tui.get_front_buffer(), 20, 15, Location::caller());
 }
+
+// https://github.com/osa1/tiny/issues/355
+#[test]
+fn test_clear_resize_recalc_scroll() {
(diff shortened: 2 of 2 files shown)


crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -179,11 +179,14 @@ impl MsgArea {
     }
 
     /// Recalculate the scroll offset due to resizing of the window
-    fn recalculate_scroll(&mut self, old_height: i32, old_total_lines: i32) {
-        if self.scroll > 0 {
-            let ratio = (self.scroll as f32 + old_height as f32) / old_total_lines as f32;
-            let total_lines = self.update_total_visible_lines();
-            self.scroll = max(0, ((ratio * total_lines as f32) as i32) - self.height);
+    fn recalculate_scroll(&mut self, old_height: i32, old_total_lines: Option<i32>) {
```
osa1, line 182: This function cannot handle `None` case, so if you keep the type as before it will become evident from the type that this really needs a `old_total_lines` argument. Please revert this and call this function only when `old_total_line` is available. It will be less surprising when you look at this code in a few months.

osa1 (commented):
This obviously fixes the panic as there isn't an `unwrap` anymore, but I'm not sure that this is the right fix. Right after `/clear` we know that the total height of rendered lines is 0. So we could reset `total_height` as `Some(0)` in `clear`, which is probably something we should do regardless of this bug, but it should also fix the bug.

I'm also a bit confused about the `lines_height` field. It seems like if I call `resize` twice without a `draw` in between that would also crash with the `unwrap`, right? I think it would be better (avoid crashes in all cases) if we have an accessor function for `lines_height` that calculates it when needed, effectively making it a cached value like the other cached values we have in e.g. messaging area lines. When it's `None`, we calculate the value, and return it. Otherwise return the existing value. No `unwrap` needed. WDYT?

crates/libtiny_tui/src/msg_area/mod.rs
trevarj: > So we could reset total_height as Some(0) in clear, which is probably something we should do regardless of this bug, but it should also fix the bug.

This is right.  Sorry

crates/libtiny_tui/src/msg_area/mod.rs
```diff
@@ -74,7 +74,9 @@ impl MsgArea {
         self.lines_height = None;
 
         self.update_total_visible_lines();
-        self.recalculate_scroll(old_height, old_total_lines.unwrap());
+        if let Some(old_total_lines) = old_total_lines {
+            self.recalculate_scroll(old_height, old_total_lines);
+        }
```
osa1, line 79: Let's also revert this. As mentioned in my other comment, this code is currently buggy and removing the `unwrap` will hide the bug. We should fix this in another PR.

osa1:
Thanks @trevarj . Did you check that this fixes the issue? I thought it should fix it, but haven't tested it myself.

Would be good to add a test also, if possible.

trevarj:
@osa1 yeah, tested as follows:
1. go to tab with a bunch of lines
2. scroll up a bit (pgup)
3. /clear
4. resize window

Added a similar unit test

osa1:
Thanks @trevarj . The test does not test the bug though. The idea is that if you revert the fix the test should fail, and with the fix it should pass. That's how you make sure it tests the bug you fixed. If I revert the fix in this PR the test still passes.

Looking at the test, I think it makes sense and I would expect it to fail without the fix. I don't understand why it passes. Maybe we misunderstand the issue?

osa1:
Ah, sorry, I was running a wrong test. OK this looks good. Will merge soon.

osa1 (approved):

PR #360, opened 2021-10-05
Refactor and fix IRC formatting handling

osa1:
This PR implements IRC formatting parsing, as described in
https://modern.ircdocs.horse/formatting.html

Previously we were only parsing the color codes (starting with `\x03`), rest of
the formatting characters were ignored.

We now parse all formatting characters, but formatting characters other than
two-digit color codes, "reverse" (reverses current background and foreground
colors), and "reset" (resets formatting to default) are currently ignored.
Handling those is left for another PR.

I also refactored mapping IRC colors to termbox colors. Previously, for some
colors, we used ANSI terminal colors in range 0-15, but those colors change
depending on the terminal color scheme (not sure if this is part of the ANSI
standard, or just something terminals do to implement color schemes). So we now
map IRC color codes to ANSI colors in range 16-255.

### DIFF
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -34,10 +34,10 @@ fn default_max_nick_length() -> usize {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Style {
-    /// Termbox fg.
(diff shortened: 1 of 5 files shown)


osa1:
cc @Kabouik @trevarj

PR #363, opened 2021-11-14
Added support for SASL EXTERNAL 

### LINKED ISSUE #196: SASL External
Can you add support for the SASL external (rfc 4422)?

Refs:
https://tools.ietf.org/html/rfc4422
https://ircv3.net/specs/extensions/sasl-3.1

trevarj:
A user can now generate a x509 certificate, register it with a server,
and provide the PEM file to tiny for use over TLS.

Closes #196

Overview:
1. Bumped rustls crates
2. Added configuration for SASL EXTERNAL, where a user specifies a path to a PEM file with a certificate inside (generated with instructions from server).
3. Use user provided certificate in TLS connector config for client authorization (crates/libtiny_client/src/stream.rs)

### DIFF
--- a/crates/libtiny_client/src/lib.rs
+++ b/crates/libtiny_client/src/lib.rs
@@ -68,16 +68,14 @@ pub struct ServerInfo {
 /// - <https://www.alphachat.net/sasl.xhtml>
 #[derive(Debug, Clone)]
 pub enum SASLAuth {
-    Plain { username: String, password: String },
-    External(SASLExternal),
-}
-
-#[derive(Debug, Clone)]
-pub struct SASLExternal {
-    /// DER-encoded X.509 certificate used for TLS Client Authorization
-    pub cert: Vec<u8>,
-    /// DER-encoded Private Key that was used to generate `cert`
-    pub key: Vec<u8>,
+    Plain {
+        username: String,
+        password: String,
+    },
+    External {
+        /// PEM-encoded X509 private cert and private key file
+        pem: Vec<u8>,
+    },
 }
 
 /// IRC client events. Returned by `Client` to the users via a channel.
@@ -397,19 +395,17 @@ async fn main_loop(
         // Establish TCP connection to the server
         //
 
-        let sasl_ext = server_info.sasl_auth.as_ref().and_then(|s| {
-            if let SASLAuth::External(s) = s {
-                Some(s)
-            } else {
-                None
-            }
-        });
+        let sasl_pem = if let Some(SASLAuth::External { pem }) = &server_info.sasl_auth {
+            Some(pem)
+        } else {
+            None
+        };
 
         let stream = match try_connect(
             addrs,
             &serv_name,
             server_info.tls,
-            sasl_ext,
+            sasl_pem,
             &mut rcv_cmd,
             &mut snd_ev,
         )
@@ -642,7 +638,7 @@ async fn try_connect<S: StreamExt<Item = Cmd> + Unpin>(
     addrs: Vec<SocketAddr>,
     serv_name: &str,
     use_tls: bool,
-    sasl: Option<&SASLExternal>,
+    sasl_pem: Option<&Vec<u8>>,
     rcv_cmd: &mut S,
     snd_ev: &mut mpsc::Sender<Event>,
 ) -> TaskResult<Option<Stream>> {
@@ -650,7 +646,7 @@ async fn try_connect<S: StreamExt<Item = Cmd> + Unpin>(
         for addr in addrs {
             snd_ev.send(Event::Connecting(addr)).await.unwrap();
             let mb_stream = if use_tls {
-                Stream::new_tls(addr, serv_name, sasl).await
+                Stream::new_tls(addr, serv_name, sasl_pem).await
             } else {
                 Stream::new_tcp(addr).await
             };
--- a/crates/libtiny_client/src/lib.rs
+++ b/crates/libtiny_client/src/lib.rs
@@ -64,8 +64,8 @@ pub struct ServerInfo {
 }
 
 /// SASL authentication mechanisms
-/// https://ircv3.net/docs/sasl-mechs
-/// https://www.alphachat.net/sasl.xhtml
+/// - <https://ircv3.net/docs/sasl-mechs>
+/// - <https://www.alphachat.net/sasl.xhtml>
 #[derive(Debug, Clone)]
 pub enum SASLAuth {
     Plain { username: String, password: String },
--- a/crates/libtiny_client/src/lib.rs
+++ b/crates/libtiny_client/src/lib.rs
@@ -64,6 +64,8 @@ pub struct ServerInfo {
 }
 
 /// SASL authentication mechanisms
+/// https://ircv3.net/docs/sasl-mechs
+/// https://www.alphachat.net/sasl.xhtml
 #[derive(Debug, Clone)]
 pub enum SASLAuth {
     Plain { username: String, password: String },
--- a/crates/libtiny_client/src/lib.rs
+++ b/crates/libtiny_client/src/lib.rs
@@ -63,11 +63,19 @@ pub struct ServerInfo {
     pub sasl_auth: Option<SASLAuth>,
 }
 
-/// SASL authentication credentials
+/// SASL authentication mechanisms
 #[derive(Debug, Clone)]
-pub struct SASLAuth {
-    pub username: String,
-    pub password: String,
+pub enum SASLAuth {
+    Plain { username: String, password: String },
+    External(SASLExternal),
+}
+
+#[derive(Debug, Clone)]
+pub struct SASLExternal {
+    /// DER-encoded X.509 certificate used for TLS Client Authorization
+    pub cert: Vec<u8>,
+    /// DER-encoded Private Key that was used to generate `cert`
+    pub key: Vec<u8>,
 }
 
 /// IRC client events. Returned by `Client` to the users via a channel.
@@ -387,10 +395,19 @@ async fn main_loop(
         // Establish TCP connection to the server
         //
 
+        let sasl_ext = server_info.sasl_auth.as_ref().and_then(|s| {
+            if let SASLAuth::External(s) = s {
+                Some(s)
+            } else {
+                None
+            }
+        });
+
         let stream = match try_connect(
             addrs,
             &serv_name,
             server_info.tls,
+            sasl_ext,
             &mut rcv_cmd,
             &mut snd_ev,
         )
@@ -623,6 +640,7 @@ async fn try_connect<S: StreamExt<Item = Cmd> + Unpin>(
     addrs: Vec<SocketAddr>,
     serv_name: &str,
     use_tls: bool,
+    sasl: Option<&SASLExternal>,
     rcv_cmd: &mut S,
     snd_ev: &mut mpsc::Sender<Event>,
 ) -> TaskResult<Option<Stream>> {
@@ -630,7 +648,7 @@ async fn try_connect<S: StreamExt<Item = Cmd> + Unpin>(
         for addr in addrs {
             snd_ev.send(Event::Connecting(addr)).await.unwrap();
             let mb_stream = if use_tls {
-                Stream::new_tls(addr, serv_name).await
+                Stream::new_tls(addr, serv_name, sasl).await
             } else {
                 Stream::new_tcp(addr).await
             };
--- a/crates/libtiny_client/src/state.rs
+++ b/crates/libtiny_client/src/state.rs
@@ -655,6 +655,7 @@ impl StateInner {
                 }
             }
 
+            // https://ircv3.net/specs/extensions/sasl-3.1.html
             AUTHENTICATE { ref param } => {
                 if param.as_str() == "+" {
                     // Empty AUTHENTICATE response; server accepted the specified SASL mechanism
@@ -664,6 +665,7 @@ impl StateInner {
                                 let msg = format!("{}\x00{}\x00{}", username, username, password);
                                 base64::encode(&msg)
                             }
+                            // Reply with an empty response (Empty responses are sent as "AUTHENTICATE +")
                             SASLAuth::External { .. } => "+".to_string(),
                         };
                         snd_irc_msg.try_send(wire::authenticate(&msg)).unwrap();
--- a/crates/libtiny_client/src/state.rs
+++ b/crates/libtiny_client/src/state.rs
@@ -633,7 +633,7 @@ impl StateInner {
                             if let Some(sasl) = &self.server_info.sasl_auth {
                                 let msg = match sasl {
                                     SASLAuth::Plain { .. } => "PLAIN",
-                                    SASLAuth::External(_) => "EXTERNAL",
+                                    SASLAuth::External { .. } => "EXTERNAL",
                                 };
                                 snd_irc_msg.try_send(wire::authenticate(msg)).unwrap();
                             } else {
--- a/crates/libtiny_client/src/state.rs
+++ b/crates/libtiny_client/src/state.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::zero_prefixed_literal)]
 
-use crate::utils;
+use crate::{utils, SASLAuth};
 use crate::{Cmd, Event, ServerInfo};
 use libtiny_common::{ChanName, ChanNameRef};
 use libtiny_wire as wire;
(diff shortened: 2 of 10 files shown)


trevarj:
Native TLS using openssl is waiting on https://github.com/sfackler/rust-native-tls/pull/209

crates/tiny/src/main.rs
```diff
@@ -142,10 +144,26 @@ fn run(
         for server in servers.iter().cloned() {
             tui.new_server_tab(&server.addr, server.alias);
 
+            let tls = server.tls;
+            let sasl_auth = if let Some(sasl_auth) = &server.sasl_auth {
```
trevarj, line 148: How can I simplify this? 🤔

trevarj: Better or worse?
```
let sasl_auth = server
    .sasl_auth
    .as_ref()
    .map(|sasl| {
        sasl_from_config(tls, sasl)
            .map_err(|e| {
                tui.add_client_err_msg(
                    &format!("SASL error for server [{}]: {}", server.addr, e),
                    &MsgTarget::Server { serv: "mentions" },
                );
            })
            .ok()
    })
    .flatten();
```

osa1: To me current version doesn't look too bad. I don't like the `flatten()` in the new version, it becomes difficult to read as there are multiple `flatten` methods (for `Option`, `Result`, `Iterator`), and you need to figure out the types to understand the code.

Another alternative:

```rust
let sasl_auth =
    server
        .sasl_auth
        .map(|sasl_auth| match sasl_from_config(tls, &sasl_auth) {
            Ok(sasl) => Some(sasl),
            Err(e) => {
                tui.add_client_err_msg(
                    &format!("SASL error for server [{}]: {}", server.addr, e),
                    &MsgTarget::Server { serv: "mentions" },
                );
                None
            }
        });
```

crates/tiny/src/main.rs
```diff
@@ -178,3 +193,41 @@ fn run(
 
     runtime.block_on(local);
 }
+
+// Helper to parse SASL config into a libtiny_client SASL struct
+fn sasl_from_config(tls: bool, sasl_config: &config::SASLAuth) -> Result<SASLAuth, String> {
+    match sasl_config {
+        config::SASLAuth::Plain { username, password } => Ok(SASLAuth::Plain {
+            username: username.to_string(),
+            password: password.to_string(),
```
osa1, line 202:
```suggestion
            username: username.clone(),
            password: password.clone(),
```

`to_string` reads like you're converting some type other than string/str to `String`, but this is cloning a borrowed `String`. `clone` makes this more clear IMO.

crates/tiny/src/main.rs
```diff
@@ -178,3 +193,41 @@ fn run(
...
+        }),
+        config::SASLAuth::External { pem } => {
+            // TLS must be on for EXTERNAL
+            if !tls {
+                Err("TLS not enabled".to_string())
```
osa1, line 207: Since we show this error message in the mentions tab, it should mention the server name, otherwise it may be difficult for the user to see where it's coming from and which server config needs to be updated. Maybe something like `format!("SASL EXTERNAL authentication only works with TLS connections. Enable TLS for {} in the config file.", server_name)`.

Just realized while typing this command, we can show errors there in the server tabs instead of mentions tab. Doesn't hurt to improve this message though. Maybe something like `"TLS is not enabled for this server, but SASL EXTERNAL authentication requires SASL. Please enable SASL for this server in the config file."` (feel free to reword).

crates/tiny/src/main.rs
```diff
@@ -142,10 +144,26 @@ fn run(
...
             tui.new_server_tab(&server.addr, server.alias);
 
+            let tls = server.tls;
+            let sasl_auth = if let Some(sasl_auth) = &server.sasl_auth {
+                match sasl_from_config(tls, sasl_auth) {
+                    Ok(sasl) => Some(sasl),
+                    Err(e) => {
+                        tui.add_client_err_msg(
+                            &format!("SASL error for server [{}]: {}", server.addr, e),
+                            &MsgTarget::Server { serv: "mentions" },
```
osa1, line 154: Should we show these errors in the server tabs? Line 145 creates the server tab.

crates/tiny/src/main.rs
```diff
@@ -178,3 +193,41 @@ fn run(
...
+            if !tls {
+                Err("TLS not enabled".to_string())
+            } else {
+                // load in a cert and private key for TLS client auth
+                match File::open(pem) {
+                    Ok(file) => {
+                        let mut buf = BufReader::new(file);
+                        // extract certificate
+                        let cert = rustls_pemfile::certs(&mut buf)
+                            .map_err(|e| format!("Could not parse pkcs8 PEM: {}", e))?
```
osa1, line 215:
```suggestion
                            .map_err(|e| format!("Could not parse PKCS8 PEM: {}", e))?
```

AFAIU, PKCS is an abbreviation for "Public-Key Cryptography Standards" (https://en.wikipedia.org/wiki/PKCS_8) so this should be in all caps.

Apparently it's also written as "PKCS <span>#</span>8" sometimes. Not sure about "PKCS8" vs. "PKCS <span>#</span>8".

crates/tiny/src/main.rs
```diff
@@ -178,3 +193,41 @@ fn run(
...
+                            .pop()
+                            .ok_or("Cert PEM must have one cert")?;
```
osa1, line 217:
```suggestion
                            .ok_or("Cert PEM must have at least one cert")?;
```

Since you use `pop` here it's OK to have more than one.

crates/tiny/src/main.rs
```diff
@@ -178,3 +193,41 @@ fn run(
...
+
+                        // extract private key
+                        buf.seek(SeekFrom::Start(0)).unwrap();
+                        let key = rustls_pemfile::pkcs8_private_keys(&mut buf)
+                            .map_err(|e| format!("Could not parse pkcs8 PEM: {}", e))?
```
osa1, line 222:
```suggestion
                            .map_err(|e| format!("Could not parse PKCS8 PEM: {}", e))?
```

crates/tiny/config.yml
```diff
@@ -16,14 +16,20 @@ servers:
...
 
       # SASL authentication
       # sasl:
-      #   username: 'tiny_user'
-      #   password: 'hunter2'
+      #   plain:
+      #     username: 'tiny_user'
+      #     password: 'hunter2'
+      #   external:
+      #     pem: "/home/.config/tiny/oftc.pem"
```
osa1, line 32: This format will break user configs. Why not parse this to the right variant depending on the available keys? If `username` and `password` are the keys, then parse to `SASLAuth::Plain`. Otherwise parse to `SASLAuth::External`.

This will be more work (we can't derive `Deserialize` anymore), but we shouldn't avoid breaking user config to avoid a few minutes of work. I'm happy to help with implementing this if you (understandably) don't want to implement `Deserialize` for this type :-)

osa1 (commented):
Thanks @trevarj! Reviewed the front-end parts for now. I'll review the client parts tomorrow.

crates/tiny/src/main.rs
trevarj: Because we want to report the error in the server tab I'm going to leave the original since it avoids the partial borrow, which can only be prevented by moving the initialization of `sasl_auth` above `tui.new_server_tab()` (takes ownership of `server.alias`).

crates/tiny/config.yml
trevarj: Actually we can just do `#[serde(untagged)]` and that should resolve which variant is configured!

crates/libtiny_client/src/lib.rs
```diff
@@ -63,11 +63,21 @@ pub struct ServerInfo {
     pub sasl_auth: Option<SASLAuth>,
 }
 
-/// SASL authentication credentials
+/// SASL authentication mechanisms
+/// https://ircv3.net/docs/sasl-mechs
+/// https://www.alphachat.net/sasl.xhtml
```
osa1, line 68:
```suggestion
/// SASL authentication mechanisms
/// - <https://ircv3.net/docs/sasl-mechs>
/// - <https://www.alphachat.net/sasl.xhtml>
```

Just to make it render nicely.

crates/libtiny_client/src/state.rs
```diff
@@ -643,15 +651,15 @@ impl StateInner {
...
-                        );
-                        snd_irc_msg
-                            .try_send(wire::authenticate(&base64::encode(&msg)))
-                            .unwrap();
+                        let msg = match auth {
+                            SASLAuth::Plain { username, password } => {
+                                let msg = format!("{}\x00{}\x00{}", username, username, password);
+                                base64::encode(&msg)
+                            }
+                            SASLAuth::External { .. } => "+".to_string(),
```
osa1, line 660: Where is this part documented? It would be good to link here to a documentation, I found it a bit hard to find.

crates/tiny/config.yml
```diff
@@ -21,9 +21,18 @@ servers:
...
+      # Providing a path to a PEM file will configure SASL EXTERNAL
+      # For SASL EXTERNAL certificate and fingerprint generation, see server documentation.
+      # You will need to register the cert's fingerprint with NickServ
+      # ex. https://www.oftc.net/NickServ/CertFP/
+      
       # sasl:
       #   username: 'tiny_user'
       #   password: 'hunter2'
+      # OR
+      #   pem: "/home/.config/tiny/oftc.pem"
```
osa1, line 35:
```suggestion
      #   pem: "/home/user/.config/tiny/oftc.pem"
```

crates/tiny/src/main.rs
```diff
@@ -178,3 +193,41 @@ fn run(
...
+fn sasl_from_config(tls: bool, sasl_config: &config::SASLAuth) -> Result<SASLAuth, String> {
+    match sasl_config {
+        config::SASLAuth::Plain { username, password } => Ok(SASLAuth::Plain {
+            username: username.clone(),
+            password: password.clone(),
+        }),
+        config::SASLAuth::External { pem } => {
+            // TLS must be on for EXTERNAL
+            if !tls {
+                Err("TLS is not enabled for this server, but SASL EXTERNAL authentication requires SASL. Please enable SASL for this server in the config file.".to_string())
```
osa1, line 207: Could you split this into multiple lines with `\` please? Example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=93a542110e8ce487180bb770e4af1b84

osa1 (approved):
This looks great, thanks again @trevarj.

Are you currently using this? I didn't test it myself.

trevarj:
> Are you currently using this? I didn't test it myself.

Yes, I think I will continue once merged too.

crates/tiny/src/main.rs
trevarj: Odd that rustfmt doesn't take care of this. Do we need to use a custom `rustfmt.toml`?

crates/libtiny_client/src/state.rs
trevarj: https://github.com/irssi/irssi/blob/master/src/irc/core/sasl.c#L205

..."the code is self-documenting" 😵‍💫

trevarj: Actually, I found it here: https://ircv3.net/specs/extensions/sasl-3.1.html

> Empty (zero-length) responses are sent as AUTHENTICATE +.

trevarj: Added in latest commit.

crates/libtiny_client/src/stream.rs
```diff
@@ -14,22 +14,53 @@ use tokio_native_tls::TlsStream;
...
-    static ref TLS_CONNECTOR: tokio_native_tls::TlsConnector =
-        tokio_native_tls::TlsConnector::from(native_tls::TlsConnector::builder().build().unwrap());
+    static ref TLS_CONNECTOR: tokio_native_tls::TlsConnector = tls_connector(None);
+}
+
+#[cfg(feature = "tls-native")]
+fn tls_connector(sasl: Option<&SASLExternal>) -> tokio_native_tls::TlsConnector {
+    let mut builder = native_tls::TlsConnector::builder();
+    if let Some(SASLExternal { cert, key }) = sasl {
+        todo!("Waiting for https://github.com/sfackler/rust-native-tls/pull/209")
```
osa1, line 28: Hm, I missed this part. We should wait for this PR, right? Would be good to mention this kind of thing as the first thing in the PR description as I often don't review line by line and miss things.

trevarj: Yes, sorry about that. Let hold until it's resolved. I suppose we _could_ merge (without a release) and then I will address the `todo!()` in another PR once the issue is resolved, but it's fine to just wait.

trevarj: This is completed now

trevarj:
Fixed conflicts after pass cmd feature.

Needs testing still.

trevarj:
@osa1 Tested using new password commands and SASL EXTERNAL. Both working as expected!
If you still want to merge this, it should be a fairly quick code review.

crates/tiny/src/main.rs
```diff
@@ -183,3 +198,30 @@ fn run(
...
+            } else {
+                // load in a cert and private key for TLS client auth
+                match File::open(pem) {
+                    Ok(mut file) => {
+                        let mut pem = Vec::new();
+                        let _ = file.read_to_end(&mut pem);
+                        Ok(SASLAuth::External { pem })
+                    }
+                    Err(e) => Err(format!("Could not open PEM file: {}", e)),
+                }
```
osa1, line 223: This block can be simplified with `std::fs::read`.

crates/tiny/src/main.rs
```diff
@@ -183,3 +198,30 @@ fn run(
 
     runtime.block_on(local);
 }
+
+// Helper to parse SASL config into a libtiny_client SASL struct
+fn sasl_from_config(tls: bool, sasl_config: &config::SASLAuth<String>) -> Result<SASLAuth, String> {
```
osa1, line 203: We could do error checking in this function in `Config::validate`, but that would mean exiting in case of an error (instead of connecting without SASL authentication). Not sure which behavior is better..

I think probably the tls checking should be done in `Config::validate` as if that check fails that means a configuration error.

crates/tiny/config.yml
```diff
@@ -21,9 +21,18 @@ servers:
       # pass: 'hunter2'
 
       # SASL authentication
+      # We currently support SASL PLAIN or SASL EXTERNAL
+      # Providing a username and password you will configure SASL PLAIN
+      # Providing a path to a PEM file will configure SASL EXTERNAL
+      # For SASL EXTERNAL certificate and fingerprint generation, see server documentation.
+      # You will need to register the cert's fingerprint with NickServ
+      # ex. https://www.oftc.net/NickServ/CertFP/
```
osa1, line 29: Remove this paragraph please. The config is getting quite large. We have an example below on how to use SASL EXTERNAL. We should add this info to the README if necessary.

osa1 (commented):
Thanks @trevarj. Code looks good. I will have to read a bit about SASL EXTERNAL before merging this. Hopefully this week..

crates/tiny/config.yml
osa1: :+1: moving this to wiki makes sense to me.

crates/tiny/src/main.rs
trevarj: Good idea. I think exiting is better. Moving it there.

trevarj: @osa1 Just to be clear, we still want to check the file read error at runtime, right? e.g display to the user in the server tab that the PEM file cannot be read.

crates/tiny/config.yml
trevarj: - [ ] Add to wiki once merged

trevarj:
> I will have to read a bit about SASL EXTERNAL 

Links that I also put in code comments:
https://ircv3.net/docs/sasl-mechs
https://www.alphachat.net/sasl.xhtml

osa1 (approved):
Thanks @trevarj. Before merging this, could you update CHANGELOG adding your name to "Thanks to ..." part and also a line to describe these changes? I would do it myself but I don't know what this is about and I don't have time to study it now. If it works I'm happy to merge.

trevarj:
@osa1 Updated CHANGELOG. I created a stub page on the wiki to add configuration details later.

Issue #359, opened 2021-09-15
Move "mentions" tab stuff to TUI


osa1:
See comment: https://github.com/osa1/tiny/pull/340#issuecomment-909071344

PR #367, opened 2021-11-28
Added welcome message to mentions tab

trevarj:
Just a "for fun" update...Open to other ideas and opinions!

image

### DIFF
--- a/crates/libtiny_tui/src/tests/mod.rs
+++ b/crates/libtiny_tui/src/tests/mod.rs
@@ -22,8 +22,8 @@ fn init_screen() {
 
     #[rustfmt::skip]
     let screen =
-        "|Any mentions to you |
-         |will be listed here.|
+        "|                    |
+         |                    |
          |                    |
          |mentions            |";
     expect_screen(screen, &tui.get_front_buffer(), 20, 4, Location::caller());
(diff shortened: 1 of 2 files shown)


trevarj:
Forgot about the tests being an issue here...I will fix them later if something like this is acceptable

osa1:
Thanks @trevarj, this looks good. A few comments:

- It looks great in the screenshot as it's centered, but on larger screens it won't be centered, and on tiny screens it will render badly. I wonder if we could easily do something like vim where the welcome message disappears as soon as you press a key that has any effect (i.e. change modes), or resize the window.

  Centering this text as the window is resized, or falling back to plain text in tiny screens will be difficult, but I wonder if we could do something like vim's welcome screen without much effort?

- We should probably mention C-n/C-p in addition to (or instead of) `alt+[1-9]` as it's more general (works when you have more than 10 tabs).

  I didn't think too much about this yet, but maybe we could show a shorter version of key bindings section in the README.

trevarj:
> but on larger screens it won't be centered, and on tiny screens it will render badly.

Definitely an issue that I thought about. I wanted to keep it like how some servers do it, and didn't want to add too much complexity.

> vim where the welcome message disappears as soon as you press a key that has any effect (i.e. change modes)

This would be really neat.

> We should probably mention C-n/C-p

Yes.

> I didn't think too much about this yet, but maybe we could show a shorter version of key bindings section in the README.

This is ideal, and I wanted to make it so that `/keys` or something opens a new tab. That is blocked by #359 , which would make it easier to just spawn dummy tabs with info. `/help` and `/names` may benefit from that too, since it will prevent flooding in a channel.

PR #384, opened 2022-03-01
Move cargo release configuration to CI

e00E:
By setting these in Cargo.toml everyone is forced to build with them. It
is better to leave the defaults to Cargo and the user building the
project. If there are important gains in binary size or speed  with
these options they can continue to be set for the official builds as
performed by CI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,9 @@ jobs:
   make_release:
     name: Make release
     runs-on: ubuntu-20.04
+    env:
+      CARGO_PROFILE_RELEASE_CODEGEN_UNITS: 1
+      CARGO_PROFILE_RELEASE_LTO: fat
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


osa1 (commented):
I'm not sure about this. I had problems with this in other projects in the past. If I'm running `make release`, `cargo build --release`, or a similar command that from the look of it seems like it should produce a release build (i.e. a build that could be distributed as a new release), then it should produce a release build. I find it frustrating when I need to set a bunch of env variables, or `touch` files etc. to generate builds for releasing.

I understand that in many projects you need to set target-specific flags or env variables etc. but in tiny this is not the case and we can keep the build steps simple, as it is the case currently.

If we still want to make this change, we should provide a script that will allow generating release builds with just one command, and use that script in CI. Devs should not have to do searching in the repo to find out how to make builds identical to the ones being released.

Secondly, at least in my workflow, I don't need `--release` to build fast. When working on tiny I use `cargo watch`, `cargo check`, or `cargo build`. Using `build` without `--release` is a better idea for development for other reasons too: it provides better backtraces, it enables debug assertions and overflow/underflow checking in arithmetic operators etc.

So I'm not sure if the motivation here is strong enough. Being able to easily build the project is important too.

PR #385, opened 2022-03-01
Move to 2021 edition

e00E:


osa1 (approved):
Thanks!

With edition 2021 closure captures are now [more precise than before](https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html), I wonder if this could simplify any code in tiny?

trevarj:
> I wonder if this could simplify any code in tiny?

Maybe running `clippy` with version 2021 will be able to detect some. I can try later.

PR #386, opened 2022-03-01
Update dependencies

e00E:
The only outdated dependency is `time` which has been completely rewritten so I did not want to update it in this PR.


osa1 (approved):
LGTM, thanks!

PR #395, opened 2022-12-10
Allow `/quit` with a message

### LINKED ISSUE #365: /quit with a message
Currently, `/quit` doesn't accept any reason for leaving, so the only way to do that is to type the QUIT message into the server tab. I think if /quit could take arguments that could be set as the reason, that would be great.

ALEX11BR:
This pull request adds an optional message parameter to the `/quit` command to be sent to the connected servers as a reason for leaving.

This pull request was tested by using a tiny and a HexChat client, both connected to the same server.

Fixes #365.

### DIFF
--- a/crates/libtiny_tui/src/tui.rs
+++ b/crates/libtiny_tui/src/tui.rs
@@ -107,8 +107,8 @@ pub(crate) enum CmdResult {
     Ok,
     /// Pass command through to cmd.rs for further handling
     Continue,
-    /// Quit command was executed
-    Quit,
+    /// Quit command was executed, with the payload as a quit message
+    Quit(Option<String>),
 }
 
 impl TUI {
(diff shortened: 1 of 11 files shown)


crates/libtiny_tui/src/tui.rs
```diff
@@ -299,7 +299,21 @@ impl TUI {
...
             }
-            Some("quit") => CmdResult::Quit,
+            Some("quit") => {
+                // Cut the "quit " part of the command
+                // (the slash is already cut outside this function),
+                // and use the rest as the message for the /quit,
+                // including other (potentially consecutive) spaces.
+                // `&cmd[("quit ".len())..]` would freeze the program
+                // if the command given has nothing more besides "/quit".
+                let reason: String = cmd.chars().skip(5 /* "quit ".len() */).collect();
```
osa1, line 309: Just use `"quit ".len()`, the compiler should be able to optimize it to 5.

I feel like the comment here is not necessary, the code is not too complicated.

Maybe we should note here that when [`SplitWhitespace::as_str`](https://doc.rust-lang.org/core/str/struct.SplitWhitespace.html#method.as_str) is stabilized we can use it to get the rest of the string (the quit message).

osa1 (approved):
Looks great, thanks!

Should we add the same parameter to `/close` as well, to be used in `quit` message when the closed tab is a server tab?

ALEX11BR:
added the same parameters to the `/close` command as well

osa1:
Thanks!

PR #397, opened 2022-12-15
Added handling for more ERR Replies

trevarj:
Moved from #344 into separate PR.
Information on what this does can be found in this comment: https://github.com/osa1/tiny/pull/344/files#r706813943

### DIFF
--- a/crates/tiny/src/conn.rs
+++ b/crates/tiny/src/conn.rs
@@ -545,6 +545,37 @@ fn handle_irc_msg(ui: &UI, client: &dyn Client, msg: wire::Msg) {
                     &format!("{} is away: {}", nick, msg),
                     &MsgTarget::User { serv, nick },
                 );
+            }
+            // ERR_BADCHANNAME
+            else if n == 479 && n_params > 2 {
+                let chan = &params[1];
+                let msg = &params[2];
+                ui.add_client_err_msg(
+                    &format!("Unable to join {}: {}", chan, msg),
+                    &MsgTarget::Chan {
+                        chan: ChanNameRef::new(chan),
+                        serv,
+                    },
+                );
+            }
+            // ERR_CHANNELISFULL    471
+            // ERR_INVITEONLYCHAN   473
+            // ERR_BANNEDFROMCHAN   474
+            // ERR_BADCHANNELKEY    475
+            // ERR_NEEDREGGEDNICK   477
+            // ERR_THROTTLE         480
+            else if (n == 471 || n == 473 || n == 474 || n == 475 || n == 477 || n == 480)
+                && n_params == 3
+            {
+                let chan = &params[1];
+                let msg = &params[2];
+                ui.add_client_err_msg(
+                    msg,
+                    &MsgTarget::Chan {
+                        chan: ChanNameRef::new(chan),
+                        serv,
+                    },
+                );
             } else {
                 match pfx {
                     Some(Server(msg_serv)) | Some(Ambiguous(msg_serv)) => {


osa1 (approved):
Thanks @trevarj!

How did you find out about ERR_BADCHANNAME, ERR_NEEDREGGEDNICK, ERR_THROTTLE?

RFC 2812 defines these as JOIN error replies:

ERR_NEEDMOREPARAMS: 461 <-- no need to handle, command parser checks number of args
ERR_BANNEDFROMCHAN: 474 <-- handled
ERR_INVITEONLYCHAN: 473 <-- handled
ERR_BADCHANNELKEY: 475 <-- handled
ERR_CHANNELISFULL: 471 <-- handled
ERR_BADCHANMASK: 476
ERR_NOSUCHCHANNEL: 403
ERR_TOOMANYCHANNELS: 405
ERR_TOOMANYTARGETS: 407
ERR_UNAVAILRESOURCE: 437

Should we handle all of them? From the names it seems like 476 and 403 may be useful. I doubt the rest will be useful.

trevarj:
> How did you find out about ERR_BADCHANNAME, ERR_NEEDREGGEDNICK, ERR_THROTTLE?

I found them all [here](https://github.com/solanum-ircd/solanum/blob/main/include/numeric.h). I believe this is the server implementation that libera uses, so I thought it would be good to cover anything it may send back to the client.

osa1:
Hmm.. Since these are not part of the standard, how do we know that another server won't use the same code for another reply?

trevarj:
> Since these are not part of the standard, how do we know that another server won't use the same code for another reply?

No idea, but you can see that some of them have RFCs and server-specific numerics documented here:
https://defs.ircdocs.horse/defs/numerics.html

osa1:
Currently all errors are handled consistently, they're shown in the server tab.

With this, some of the errors will be shown in the channel tab. This will add inconsistency that didn't exist before, because we don't (and can't) handle all of the errors.

(We can't handle all of the errors because servers use non-standard replies, as shown in the ircdocs link above)

So I argue that this actually makes the UX worse by adding an inconsistency.

The motivation for this was #344: with that PR we create a channel tab on `/join` (without waiting for the `JOIN` reply), and once we have a tab for a channel it makes sense to show errors in the channel tab.

However:

1. Some errors will still be shown in the server tab, as mentioned above (inconsistent)
2. There are other ways of setting the channel settings

I suggest in #344 we store channel settings in TUI, and when we create a tab for a channel,we use the map for channel settings to find the settings for the channel, and create the tab with that settings.

Maybe something like:

- Rename `Server.join` (in TUI config) to `Server.chan_configs`, with type `Map<ChanName, TabConfig>`.
- Option 1: when creating a tab in a server, use `chan_configs` to find the settings for the tab.
- Option 2: don't store tab settings in tab, always use `chan_configs` to get the current values of `ignore` and `notify`

Currently it seems like we're doing both of options (1) and (2): notification settings are stored in `Tab`, but ignore settings are stored in the config.

I think I prefer (2) as that means we don't duplicate information.

An implication of (2) is that if we update settings for a tab, close the tab (e.g. leave a channel), then create it again (e.g. join again), we will use the updated settings instead of the ones in the config file.

If we don't want that (i.e. we want to use the settings from the config file on leave and then join again) we should go with option (1).

@trevarj WDYT?

trevarj:
Jeez, I need some time to digest it. I haven't looked at or thought about this in over a year so I pretty much forgot everything that I had in mind before.

I only care about the way that I use this feature in my own usecase 😈  - to set the configurations on launch only. I don't actually care about what happens on rejoin, so therefore I also don't mind (2).

Maybe just close this if we will not need it in #344, and then continue the discussion over there.

osa1:
Closing this one. I'll move the relevant parts to #344 as a review comment.

Issue #364, opened 2021-11-15
Implement italic, bold, etc. rendering


osa1:
https://github.com/osa1/tiny/blob/b7094c927e4e238ba427dfda25479d24bbd62619/crates/libtiny_tui/src/msg_area/line.rs#L115-L121

We recently implemented parsing IRC format characters for bold, italic, underline, strikethrough, and monospace, but we currently ignore them. xterm and xterm-compatible terminals have control characters for these (except monospace, as all text is monospace in most terminals), so we should update termbox first to allow rendering bold, italic etc. then use it in TUI.

zschaffer:
I set up the ability for termbox to render out the incoming IRC format codes. Next up would be adding in a way for tiny users to format their own messages. Unsure what the best way to implement this on the client side is.

Could use `*word*`, `_word_` etc and then strip the format characters or have a /bold word /unbold sort of command structure.

osa1:
@zschaffer I'm also not sure what would be the best way of allowing italic, bold etc. input. I like the Markdown syntax that you show, but we should provide a way to actually send star, underscore etc. instead of making the text bold or italic.

Rendering these and allowing tiny to send formatted text are separate features and don't need to be merged together. If you implemented rendering part send a PR and I will review & merge. 🙂

trevarj:
I like the markdown syntax too, and maybe we can have a config that can be toggled with `/formatting` or something. 

Will this work with [the current standard](https://modern.ircdocs.horse/formatting.html#client-behaviour)?

osa1:
> Will this work with [the current standard](https://modern.ircdocs.horse/formatting.html#client-behaviour)?

Yes, we parse the codes listed there. ([source](https://github.com/osa1/tiny/blob/master/crates/libtiny_wire/src/formatting.rs))

PR #401, opened 2023-01-17
add formatting to termbox, render in tui (#364)

zschaffer:
added a fn for Line that updates its current style, OR'ing it with the attribute you want to set.

Had to add a few more bitmasks for Italic and Strikethrough as well.

This setup allows attributes to 'stack' easily, ie: receiving a message that's both bold and italic will show properly.

image


osa1 (approved):
Looks great, thanks!

PR #410, opened 2023-08-04
Support tiny commands in keymap config

### LINKED ISSUE #403: String input key binding
Currently, there is a key action `input: CHAR` which adds the `CHAR` to input area when triggered. However, I would like to be able to input a list of chars (aka string) so that I could use, for example, `ctrl_l` to send `/clear`, or `ctrl_d` to send `/close`. I see no way to do this with actions available at the moment, but it does not seem to be hard to implement

nate-sys:
Allows the user to specify a config of this sort:
```yaml
key_map:
  ctrl_d:
    run_command: quit
```

Fixes #403

### DIFF
--- a/crates/libtiny_tui/src/key_map.rs
+++ b/crates/libtiny_tui/src/key_map.rs
@@ -368,10 +368,9 @@ fn deser_keymap() {
         .insert(Key::Ctrl('a'), KeyAction::InputMoveCursStart);
     expect.0.insert(Key::Ctrl('e'), KeyAction::TabGoto('1'));
     expect.0.insert(Key::Char('a'), KeyAction::Input('b'));
-    expect.0.insert(
-        Key::Char('b'),
-        KeyAction::RunCommand("clear".to_string()),
-    );
+    expect
+        .0
+        .insert(Key::Char('b'), KeyAction::RunCommand("clear".to_string()));
 
     let key_map: KeyMap = serde_yaml::from_str(s).unwrap();
     assert_eq!(expect, key_map);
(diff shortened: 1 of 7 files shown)


crates/libtiny_tui/src/key_map.rs
```diff
@@ -357,13 +359,19 @@ fn deser_keymap() {
       tab_goto: 1
     a: 
       input: b
+    b:
+      run_irc_command: clear
```
osa1, line 363: I think this may be confused with actual IRC commands like PRIVMSG, QUIT, JOIN etc.

I suggest just "command".

crates/libtiny_tui/src/key_map.rs
```diff
@@ -31,6 +31,7 @@ pub(crate) enum KeyAction {
     MessagesScrollBottom,
 
     Input(char),
+    RunIrcCommand(String),
```
osa1, line 34: As mentioned in my other comment, the name is confusing. I suggest we just call this "Command".

osa1 (commented):
Thanks! Code looks great. Added some inline comments regarding naming.

crates/libtiny_tui/src/key_map.rs
nate-sys: Right. Will fix.

crates/libtiny_tui/src/key_map.rs
```diff
@@ -357,13 +359,18 @@ fn deser_keymap() {
       tab_goto: 1
     a: 
       input: b
+    b:
+      run_command: clear
```
osa1, line 363: I'd drop "run" as well as there's nothing else you can do with a command other than to run, so it's not helpful.

I'll do it after the merge.

osa1 (approved):
Thanks again!

PR #411, opened 2023-08-07
Replace clap with a simple hand-written arg parser

osa1:
clap is too heavyweight for what tiny needs. Replacing it with a simple hand-written parser reduces debug binary sizes a constant 14 MiB (!), and release binary sizes ~400 KiB.

- Default build, debug:
  - Before: 73M
  - After: 59M
  - Diff: -14M, -29%

- Default build, release:
  - Before: 6.4M
  - After: 6.0M
  - Diff: -0.4M, -6%

- Native TLS build, release:
  - Before: 4.6M
  - After: 4.2M
  - Diff: -0.4M, -8%

Also saves ~10s build time and ~3s `cargo check` time.

Error messages and the help message is copied from what clap printed and modified.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,17 +155,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1181e1e0d1fce796a03db1ae795d67167da795f9cf4a39c37589e85ef57f26d3"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ddiss:
:partying_face: thanks a lot for keeping tiny tiny

PR #412, opened 2023-08-30
Add tab_joinpart to default config

jubalh:
So that users can easily find the option to tweak it.

Related to https://github.com/osa1/tiny/issues/262 c6d20b68 introduced this option.

### DIFF
--- a/crates/tiny/config.yml
+++ b/crates/tiny/config.yml
@@ -164,6 +164,10 @@ colors:
         bg: default
         attrs: [bold]
 
+    tab_joinpart:
+        fg: 11
+        bg: default
+
 # Custom key mappings. Syntax is:
 #
 # [key_name]: [key_action]


crates/tiny/config.yml
```diff
@@ -40,6 +42,7 @@ servers:
       # sasl:
       #   username: tiny_user
       #   password: hunter2
+      #   pem: "/home/user/.config/tiny/oftc.pem"
```
osa1, line 45: Could you also clarify somewhere (maybe right above the `sasl:` line) that you either need `pem` or `username` + `password`? Currently it looks like you need all of them if you want to use SASL authentication.

osa1 (approved):
Thanks! I added a small comment.

crates/tiny/config.yml
jubalh: I tried to make this clear with `Choose only one of the options below.` in the general sasl description at the top. But maybe I should rephrase somehow..

jubalh: Choose only one of the options below. -> Choose either `username` and `password` _or_ `pem` in the section below.

crates/tiny/config.yml
```diff
@@ -29,6 +29,9 @@ servers:
       #
       # - `pass`: Sends `PASS` IRC command when connecting.
       # - `sasl`: For SASL authentication.
+      #    You can either authenticate using a password or use a certificate
+      #    with SASL EXTERNAL. Choose either `username` and `password` _or_ `pem`
+      #    in the section below.
```
osa1, line 34: Because we're talking about `sasl` section already we should drop the "in the section below" part because it may sound like there are two sections (the `sasl` section + "the section below").

Alternatively we could also move this part right above the `sasl:` line.

osa1:
Thanks again!

PR #418, opened 2023-11-10
README.md: replace freenode reference with libera

ralphptorres:
Freenode is dead. Many open-source projects left the network due to the 2021 situation (of hostile takeover, etc.) and have since migrated to Libera.


osa1 (approved):
Thanks!

PR #419, opened 2023-11-10
PKGBUILD: add openssl-1.1 dependency

ralphptorres:
tiny does not run if libssl.so.1.1 is missing. the current dependency on openssl does not seem to address this but openssl-1.1 does

### DIFF
--- a/pkg/archlinux/PKGBUILD
+++ b/pkg/archlinux/PKGBUILD
@@ -8,7 +8,7 @@ arch=('x86_64')
 provides=('tiny')
 url="https://github.com/osa1/tiny"
 license=('MIT')
-depends=('openssl' 'dbus')
+depends=('openssl' 'openssl-1.1' 'dbus')
 makedepends=('git' 'rust' 'cargo')
 source=("git+$url#commit=5e5c90c8f6b85b5ba38c974ed8113beef0e916ed") # tag: v0.9.0
 sha512sums=(SKIP)


osa1:
Without `--no-default-features --features=tls-native` (as in this `PKGBUILD`) tiny should build with `rustls` instead of `native-tls`, and `rustls` shouldn't depend on `libssl.so`.

`ldd` output with no flags (as in `PKGBUILD`):

```
$ ldd target/debug/tiny
	linux-vdso.so.1 (0x00007ffc05bdc000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9c404b6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9c3fa00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9c404f9000)
```

There's no `libssl.so` here.

`ldd` output with `--no-default-features --features=tls-native`:

```
$ ldd target/debug/tiny
	linux-vdso.so.1 (0x00007ffce45ce000)
	libssl.so.3 => /lib/x86_64-linux-gnu/libssl.so.3 (0x00007fb6dd15c000)
	libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x00007fb6dcc00000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fb6dd71b000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fb6dc800000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fb6dd75e000)
```

This one uses `libssl.so`.

Could you update the commit in the file to the latest master branch and try again?

ralphptorres:
Sorry for this erroneous pr. Apparently, I have tiny-irc-client-bin installed which fetches the -libssl-dbus variant from this repo's releases section (hence depending on libssl.so as described). Installed that because both tiny-irc-client and tiny-irc-client-git are outdated.

On another note, the script in this repo conflicts with the latest in [AUR git]. Idk which repo serves as source of truth wrt the PKGBUILD script.

[AUR git]: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=tiny-irc-client-git

PR #421, opened 2023-11-11
PKGBUILD: refactor, improve, and sync changes with the version found in AUR git

ralphptorres:
This patch mirrors the PKGBUILD from AUR git, refactors, and improves it via the following commits. These commits are already present in the AUR git so anyone can test this script by installing tiny-irc-client-git.

Context: Prior to these commits, this script was not updated since 2021/11.  I reached out to the maintainers to suggest improvements. They mentioned they no longer use tiny then gave me the blessing to maintain the package myself.

### DIFF
--- a/pkg/archlinux/PKGBUILD
+++ b/pkg/archlinux/PKGBUILD
@@ -10,8 +10,8 @@ arch=(x86_64)
 url=https://github.com/osa1/tiny
 license=(MIT)
 
-depends=(openssl dbus)
-makedepends=(git rust)
+depends=(dbus)
+makedepends=(git cargo)
 provides=(${pkgname%-git})
 conflicts=(${pkgname%-git})
 source=(git+$url)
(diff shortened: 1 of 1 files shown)


pkg/archlinux/PKGBUILD
```diff
@@ -1,25 +1,42 @@
...
+    git describe --tags --long | \
+        sed -e 's/\([^-]*-\)g/r\1/' -e 's/-/./g' -e 's/^v//'
+}
+
 build() {
-          cargo build \
-            --manifest-path "$srcdir/tiny/Cargo.toml" \
-            --release
+    cd $_pkgname
+    cargo install --path crates/$_pkgname --features=desktop-notifications
```
osa1, line 30: This invocation builds with `rustls` and doesn't depend on `openssl`. You can either remove the `openssl` dependency or run `cargo install ... --no-default-features --features=desktop-notifications,tls-native` and depend on `openssl` instead of `rustls` (`--no-default-features` is needed to disable `tls-rustls` feature which we replace with `tls-native`).

osa1 (commented):
Thanks! I've added an inline comment.

pkg/archlinux/PKGBUILD
ralphptorres: Updated with the former, thanks.

PR #423, opened 2023-11-12
Use dbus instead of zbus for desktop notifications

osa1:
zbus adds an entire async runtime to the binary and brings 90 dependencies (102 to 192).

The final release binary size is 8.2M (4.2M stripped) with zbus and 6.3M (2.8M stripped) with dbus. Statically linked, 8.5M (4.4M stripped) with zbus and 6.8M (3.1M stripped).

Since the dbus runtime dependency is fairly common (for example used by the google-chrome package on Ubuntu) I suspect most users will already have it installed, or at least will be able to easily install it.

Runtime dependencies before:

    linux-vdso.so.1
    libgcc_s.so.1
    libc.so.6
    /lib64/ld-linux-x86-64.so.2

After:

    linux-vdso.so.1
    libdbus-1.so.3
    libgcc_s.so.1
    libc.so.6
    /lib64/ld-linux-x86-64.so.2
    libsystemd.so.0
    liblzma.so.5
    libzstd.so.1
    liblz4.so.1
    libcap.so.2
    libgcrypt.so.20
    libgpg-error.so.0


osa1:
@ralphptorres could you check if this builds and runs with your changes in #421? I suspect the part `depends=(... dbus)` should be enough but I'm not sure. I don't use Arch so I can't try myself.

ralphptorres:
Can confirm that it builds and runs fine. Also found out that `dbus` and the other runtime deps above are already included in `base` (or its deps like `systemd`), which every Arch install has. So I guess we can just remove `dbus` as a dependency?

Also, do you prefer this feature flag over the `zbus` one? If so, we can update the PKGBUILD in #421.

ralphptorres:
Decided to include `dbus` as dependency despite my previous argument.

Issue #402, opened 2023-02-10
Add completion character after autocompleting a nick


trevarj:
This was reported in IRC, I am just relaying it for visibility.

Similar to hexchat or [irssi](https://irssi.org/documentation/settings/#completion-char), it would be useful for some people to have a character placed after a nick is autocompleted, such as a colon or comma (i.e `nick: reply message` or `nick, reply message`). This should be configurable and only occur when the nick is at the beginning of the line (I think).

PR #427, opened 2024-01-07
Add completion_char feature

iulianalexa:
Implements the feature described in [Issue #402](https://github.com/osa1/tiny/issues/402).
To add the completion_char, add the following line to the config:

```
completion_char: ":"
```

The character will only be appended to a completion at the start of the line.

### DIFF
--- a/crates/libtiny_tui/src/messaging.rs
+++ b/crates/libtiny_tui/src/messaging.rs
@@ -169,7 +169,8 @@ impl MessagingUI {
             }
             KeyAction::InputAutoComplete => {
                 if self.exit_dialogue.is_none() {
-                    self.input_field.autocomplete(&self.nicks, &self.completion_char);
+                    self.input_field
+                        .autocomplete(&self.nicks, &self.completion_char);
                 }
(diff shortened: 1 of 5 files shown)


crates/libtiny_tui/src/messaging.rs
```diff
@@ -39,6 +39,9 @@ pub(crate) struct MessagingUI {
 
     /// Last timestamp added to the UI.
     last_ts: Option<Timestamp>,
+
+    // Autocompletion character to be used when at the start of the line. Set with `set_completion_char`.
+    completion_char: Option<String>,
```
trevarj, line 44: I think it would make more sense to move this field into `InputArea`. Then you can avoid passing the `completion_char` into the `autocomplete()` function above, and just use `self.completion_char`.

trevarj (changes requested):

crates/libtiny_tui/src/messaging.rs
iulianalexa: Done

osa1:
IIRC this was a class assignment. Closing.

Issue #369, opened 2021-12-02
Ignoring a user?


097115:
We can't ignore a certain user, can we? :) (Oh, god, some bots are really too chatty...)

osa1:
This is currently not possible. I also needed this a few times. Marking as "feature request", would be good to have this feature.

PR #428, opened 2024-01-08
Add user blocking feature

iulianalexa:
This pull request addresses [Issue #369](https://github.com/osa1/tiny/issues/369) and implements the requested feature.

To block a user, add this line to the config:
```
blocked_users: ["~user@example"]
```

Inside the blocked_users list, add the hostmasks of the users you would like to block.

If this gets approved, we could eventually expand on the block function by implementing features such as:

- Regex matching (e.g. "~user@*")
- Separate .yml file for the blocked users
- /block, /unblock commands

### DIFF
--- a/crates/libtiny_tui/src/config.rs
+++ b/crates/libtiny_tui/src/config.rs
@@ -4,7 +4,7 @@
 use libtiny_common::{ChanName, ChanNameRef};
 use serde::de::{self, Deserializer, MapAccess, Visitor};
 use serde::Deserialize;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::Path;
 
 use termbox_simple::*;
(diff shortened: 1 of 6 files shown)


crates/libtiny_tui/src/tui.rs
```diff
@@ -108,6 +108,9 @@ pub struct TUI {
 
     /// TabConfig settings loaded from config file
     tab_configs: TabConfigs,
+
+    /// List of blocked users
+    blocked_users: Vec<String>,
```
trevarj, line 113: I think it would be better to use a `HashSet<String>` here to have the O(1) lookup

trevarj (changes requested):

crates/libtiny_tui/src/tui.rs
iulianalexa: Done

crates/libtiny_tui/src/config.rs
```diff
@@ -31,6 +31,9 @@ pub(crate) struct Config {
 
     #[serde(default)]
     pub(crate) key_map: Option<KeyMap>,
+
+    #[serde(default)]
+    pub(crate) blocked_users: HashSet<String>,
```
trevarj, line 36: You will also want to add this to the [sample config file ](https://github.com/osa1/tiny/blob/master/crates/tiny/config.yml)

iulianalexa: Done

osa1:
IIRC this was a class assignment. Closing.

PR #433, opened 2024-08-06
Upgrade ring to 0.17

hack3ric:
Bumping tokio-rustls to exactly 0.24 upgrades ring to 0.17 without introducing breaking changes, despite having deprecated methods.


osa1 (approved):
Thanks!

PR #438, opened 2024-12-19
Use IndexMap in byte_seq_parser macro

### LINKED ISSUE #437: reproducible builds problem
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds) (sponsored by the NLnet NGI0 fund), I found that
our `tiny` package varies in every build.

Another developer pointed at HashMaps in
https://github.com/osa1/tiny/blob/ee8615a55256b242b02e6f8a2350ca7f39aca517/crates/term_input_macros/src/tree.rs#L6
with non-deterministic iteration order.

Maybe it could be replaced with [indexmap](https://github.com/indexmap-rs/indexmap), [ordermap](https://github.com/indexmap-rs/ordermap) or something else...

osa1:
This make map iteration deterministic and makes tiny builds reproducible.

Fixes #437.


osa1:
The command

```
  cargo test --no-default-features \
    --features "tls-native desktop-notifications"
```

Is currently failing on `macos-latest`. It started a while ago and the run passes if I restart it a few times. We should investigate it separately.

bmwiedemann:
Thanks for the quick patch. I cannot test it easily with my tools because of the added dependency. I expect it to be fine.

PR #445, opened 2025-05-20
Untangles the keystrokes ctrl-h and backspace

nuvw:
Untangles the keystrokes ctrl-h and backspace

Based on our discussion in #444, this seems to me like an obvious and simple change. But feel free to ignore it if you have any doubts (in particular since at some point in the past you felt the need to map the two keystrokes to the same `Key`; maybe there was a good reason for it after all).

Effectively, as I see it, this adds the possibility to define separate key bindings for `ctrl_h` and `backspace` in the config file, while not changing the default behavior of tiny at all.

Of course, there is now also the question of what action `ctrl-h` should be mapped to by default. But it seems to be standard to have it do the same as `backspace`, namely to delete the character before the cursor. At least I have checked that this is also the default behavior of vim, helix and the fish shell.

If you accept this PR, then I am not so sure about #444 anymore since (as you also point out in the discussion there) it taints `libtiny_tui/src/key_map.rs` with low-level knowledge about equality of certain key combinations for very little (and now even less) gain.


osa1:
Thanks!

> If you accept this PR, then I am not so sure about https://github.com/osa1/tiny/pull/444 anymore since (as you also point out in the discussion there) it taints libtiny_tui/src/key_map.rs with low-level knowledge about equality of certain key combinations for very little (and now even less) gain.

We can merge that as well, but we should update it now removing the `ctrl-h` part, and add comments to document why the special cases for `i` and `[` are needed.

We should also not forget to update the CHANGELOG with the changes.

PR #448, opened 2025-09-05
Add 'user' option for server

anarsoul:
ZNC uses 'user' as a way to specify what network to connect to if there are several networks configured. E.g. myspecialuser/oftc or myspecialuser/libera

### DIFF
--- a/crates/libtiny_client/src/state.rs
+++ b/crates/libtiny_client/src/state.rs
@@ -254,8 +254,9 @@ impl StateInner {
         snd_irc_msg
             .try_send(wire::nick(&self.current_nick))
             .unwrap();
+        let user = self.server_info.user.as_ref().unwrap_or(&self.nicks[0]);
         snd_irc_msg
-            .try_send(wire::user(&self.nicks[0], &self.server_info.realname))
+            .try_send(wire::user(user, &self.server_info.realname))
             .unwrap();
     }
 
--- a/crates/libtiny_wire/src/lib.rs
+++ b/crates/libtiny_wire/src/lib.rs
@@ -23,8 +23,8 @@ pub fn quit(reason: Option<String>) -> String {
     }
 }
 
-pub fn user(hostname: &str, realname: &str) -> String {
-    format!("USER {hostname} 8 * :{realname}\r\n")
+pub fn user(user: &str, realname: &str) -> String {
+    format!("USER {user} 8 * :{realname}\r\n")
 }
 
 pub fn nick(arg: &str) -> String {
--- a/crates/tiny/src/config.rs
+++ b/crates/tiny/src/config.rs
@@ -59,6 +59,11 @@ pub(crate) struct Server<P> {
     #[serde(default)]
     pub(crate) pass: Option<P>,
 
+    /// User name to be used in connection registration
+    /// If it is not specified, the first nick will be used instead
+    #[serde(default)]
+    pub(crate) user: Option<String>,
+
     /// Real name to be used in connection registration
     #[serde(deserialize_with = "deser_trimmed_str")]
     pub(crate) realname: String,
@@ -319,6 +324,7 @@ impl Config<PassOrCmd> {
                 port,
                 tls,
                 pass,
+                user,
                 realname,
                 nicks,
                 join,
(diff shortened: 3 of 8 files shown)


crates/libtiny_wire/src/lib.rs
```diff
@@ -23,8 +23,8 @@ pub fn quit(reason: Option<String>) -> String {
     }
 }
 
-pub fn user(hostname: &str, realname: &str) -> String {
-    format!("USER {hostname} 8 * :{realname}\r\n")
+pub fn user(user: &str, hostname: &str, realname: &str) -> String {
+    format!("USER {user} {hostname} * :{realname}\r\n")
```
osa1, line 27: This format doesn't seem right according to RFC 2812: https://github.com/osa1/tiny/blob/c5e68d9ff10ba2144d0f6c3389db5de1f3e0e678/rfcs/rfc2812.txt#L577-L609

According to the RFC, `USER` message should have 4 parameters:

- User: this is currently the `hostname` argument
- Mode: this is currently hard-coded as `8`. I don't remember what this means, we should probably document.
- Unused: hard-coded as `*`, this is copied from the RFC.
- Realname: the `realname` argument.

With this you make the `mode` parameter `hostname`, which doesn't seem right according to the RFC. RFC says "mode parameter should be numaric, and can be used to automatically set user modes ...".

Maybe you just want to pass different `hostname` argument to this function? Did you check if the `hostname` is configurable in the config file? (I don't remember the answer, it may already be configurable)

osa1 (commented):
Hey @anarsoul, thanks for the patch.

I think it's not quite right, see my inline comment.

There are also CI failures.

crates/libtiny_wire/src/lib.rs
anarsoul: Ugh, I referred to https://www.rfc-editor.org/rfc/rfc1459#section-4.1.3

`hostname` isn't configurable, current code just uses `&self.nicks[0]` for `hostname`. But anyway, the name is misleading, as per RFC2812 it should be `user`

I'll fix it up tonight.

osa1: In that case it might make sense to follow one of the more recent specs. Maybe https://modern.ircdocs.horse/#user-message.

According to this spec there should be one user name and one real name. So the hostname can be dropped.

Could you try it and see if it works on some of the popular servers?

anarsoul: Done, I updated the patch to just replace `hostname` with `user` and to allow specifying it in the config.

I confirmed that the change works for me on OFTC and Libera and that it works correctly with ZNC as well.

crates/libtiny_client/src/state.rs
```diff
@@ -254,9 +254,15 @@ impl StateInner {
...
-            .unwrap();
+        if let Some(ref user) = self.server_info.user {
+            snd_irc_msg
+                .try_send(wire::user(user, &self.server_info.realname))
+                .unwrap();
+        } else {
+            snd_irc_msg
+                .try_send(wire::user(&self.nicks[0], &self.server_info.realname))
+                .unwrap();
+        }
```
osa1, line 265: You could simplify this by having a local (not tested)
```rust
let user = self.server_info.user.as_ref().unwrap_or(&self.nicks[0]);
snd_irc_msg.try_send(wire::user(user, ...)).unwrap();
```

crates/tiny/src/config.rs
```diff
@@ -59,6 +59,10 @@ pub(crate) struct Server<P> {
     #[serde(default)]
     pub(crate) pass: Option<P>,
 
+    /// User name to be used in connection registration
```
osa1, line 62: Please also mention what user name is used when this is not available.

osa1 (approved):
Thanks @anarsoul , looks great.

Just one more thing -- could you update the default config file in `crates/tiny/config.yml` with the new field, with documentation? I think we should add this field right above `realname`:

```
# Servers to automatically connect.
servers:
    - addr: irc.oftc.net
      port: 6697
      tls: true
      # HERE
      realname: yourname
      nicks: [tiny_user]
```

To be backwards compatible with the current version it should be commented-out, with documentation.

PR #449, opened 2025-10-15
Packaged for the Debian project

Nilsonfsilva:

PR #450, opened 2025-10-16
Added man page for tiny to the Debian project

Nilsonfsilva:
Hi!
As requested, here is the breakdown of PR 449.

Feel free to incorporate the man page into your project. Maintenance may bring you another request.

The issue with creating a man page is that
Debian's packaging system complains when a binary is generated and doesn't have a man page. To address this, we created one.

Upstream submission is available if they want to incorporate it into their project.


osa1:
As mentioned in #449, if we want a man page, the man page and at least some of the README should be generated from a single file, to avoid copy/pasting same stuff in two files.

PR #453, opened 2025-10-16
Apply Debian patch 001.relax-version-depends from 0.13.0

Nilsonfsilva:
Hi!
follow the other patch where Cargo.toml is updated.

### DIFF
--- a/crates/libtiny_client/Cargo.toml
+++ b/crates/libtiny_client/Cargo.toml
@@ -11,15 +11,15 @@ tls-native = ["native-tls", "tokio-native-tls"]
 tls-rustls = ["rustls-native-certs", "tokio-rustls", "rustls-pemfile"]
 
 [dependencies]
-base64 = "0.21"
+base64 = ">=0.21"
 futures-util = { version = "0.3", default-features = false }
 lazy_static = "1.4"
 libtiny_common = { path = "../libtiny_common" }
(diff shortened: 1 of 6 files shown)


Nilsonfsilva:
Hi
You can maintain it in the most appropriate format in tiny's Cargo.toml.

On the Debian side, we're always adjusting it to make the build work.

What I did was adjust it to make the build work.

Debian always rebuilds packages to check if its ecosystem is working correctly.
When something breaks, the Rust team, which I'm part of, steps in to fix it.

Don't worry about that.

Maintain Cargo.toml in whatever format you see fit.



Nilson F. Silva


________________________________
De: Ömer Sinan Ağacan ***@***.***>
Enviado: sábado, 18 de outubro de 2025 04:39
Para: osa1/tiny ***@***.***>
Cc: Nilsonfsilva ***@***.***>; Author ***@***.***>
Assunto: Re: [osa1/tiny] Apply Debian patch 001.relax-version-depends from 0.13.0 (PR #453)


@osa1 commented on this pull request.

We can't use >= with pre-1 versions because in semver versions before 1.0 doesn't have any backwards compatibility guarantees, even in minor increments.

Even if you make it build with these changes a cargo update would break the build later.

So you should find the version of the dependencies you want (the exact 0.x.y version) and use that instead.

Issue #379, opened 2022-01-19
Bug: Missing characters when pasting green square emoji


alxgnon:
I pasted this text into tiny IRC:
🟩🟩🟩🟩🟩

I got:
![Screenshot_2022-01-19_18-24-17](https://user-images.githubusercontent.com/4480082/150234446-21231368-c0d9-4670-968b-e4db3d1eb179.png)

For some reason there is only 3 squares instead of 5.

trevarj:
Ah yes, all multi-width (multi-cell) characters are not rendered properly in the tui. I can take a look at this.

trevarj:
Leaving this here for later: https://github.com/ridiculousfish/widecharwidth (see generated Rust code).

We can use it to get widths of wide characters (such as 🟩).

osa1:
Multi-width rendering in the message area should be supported, we use `unicode_width` for that: https://github.com/osa1/tiny/blob/5bcba53d4dc9fc0cb63e5efb635a815e26eff05d/crates/termbox/src/lib.rs#L275

Without any investigation, I'm guessing this could be a bug in the input area.

osa1:
Related: https://github.com/osa1/tiny/issues/306

alxgnon:
> Without any investigation, I'm guessing this could be a bug in the input area.

I've seen this bug in the chat log as well. Basically anywhere that there would be 2 consecutive squares, only 1 of them would show up.

Also, if I input one square emoji, and a letter afterwards, the square emoji will "swallow" the letter, in the input field and also in the chat log.

PR #464, opened 2026-03-18
fix(tui): resolve CJK character display and clipping issues

y1feng200156:
Root Cause:

- Characters were assumed to have a fixed width of 1 during rendering, whereas CJK characters occupy 2 columns.

- This caused CJK characters to overlap subsequent columns, leading to corrupted display.

Changes:

- termbox.rs: Updated print_chars to use UnicodeWidthChar for accurate width calculation.

- tab.rs: Integrated UnicodeWidthChar into Tab rendering.

- msg_area/line.rs: Updated Line rendering logic with UnicodeWidthChar.

- input_line.rs: Applied UnicodeWidthChar to the input area rendering.

- line_split.rs: Verified that line height calculation already correctly handles Unicode width (no changes needed).

All tests passed. Build successful.


osa1 (commented):
From a quick look the code looks fine, but please remove the extra files (added by your AI agent I assume) and add tests.

Could please confirm if this fixes #379.

I'll have another look after adding the tests and the tests pass. I don't want to debug slop myself. See the existing tests in `crates/libtiny_tui/src/tests/...`.

osa1 (commented):
Thanks. The tests are good but I think they may not be sufficient. We should also check that the TUI deals with character widths properly when scrolling, and a line is cut off or split into multiple lines.

For example, if I create a TUI (in the tests) with width 10, I can fit 10 ASCII characters:

```
|1234567890|
```

Now if I fill the screen with the square characters, I can fit 5 of them, and when I check the buffer contents I should get something like:

```
|🟩🟩🟩🟩🟩|
```

instead of

```
|🟩🟩🟩🟩🟩     |
```

(The incorrect version has 5 spaces after the square characters)

Then, in the correct version, if I add an ASCII character like '1', it should look like this in the input line:

```
|🟩🟩🟩🟩1 |
```

Note that this has one space on the right side of the input line as the whole square character needs to be skipped on the left. (scroll = 1)

Then we can add another character and it should look like:

```
|🟩🟩🟩🟩12|
```

We should test similar things with splitting long lines. E.g. for the same screen size, if I have one space between each square it should look like:

```
|🟩 🟩 🟩  |
|🟩 🟩     |
```

and then I can keep resizing to test all kinds of things. E.g. if I resize it to 11:

```
|🟩 🟩 🟩 🟩|
| 🟩        |
```

(I don't remember if we add a space on the new line when we split at a space and space doesn't fit into the first line, worth checking this)

Then 12:

```
|🟩 🟩 🟩 🟩 |
|🟩          |
```

Then 13:

```
|🟩 🟩 🟩 🟩  |
|🟩           |
```

Then 14:

```
|🟩 🟩 🟩 🟩 🟩|
|              |
```

You get the idea.

Now, this is a bit tricky to check because when you get the TUI's buffer as a `String` you'll get one character per terminal grid cell, so wide characters will look like: charcter, then some number of spaces. There's no easy way to check this, I think we'll have to take this into consideration and check accordingly. E.g. a line with one square and one space will look like "🟩  " (note: 2 spaces) because that's how it'll be represented in the termbox buffer.

The code seems complicated (and not documented enough EDIT: to clarify, I'm talking about my own, old code. Not the code in this PR.) so I'll make a pass once the tests are final and working.

PR #465, opened 2026-03-21
fix(term_input): Handle IME (Input Method Editor) input correctly

y1feng200156:
## Problem

When using Chinese/Japanese/Korean IME to input characters, the ESC key followed by multi-byte UTF-8 sequences was incorrectly parsed as Alt+key combinations. This caused IME input to be misinterpreted.

For example, typing Chinese characters with IME would produce unexpected Alt+char events instead of the intended characters.

## Solution

Detect multi-byte UTF-8 sequences (bytes >= 0xC0) after ESC and treat them as IME input rather than Alt+key. Return ESC as a separate event and let the UTF-8 bytes be parsed separately as regular characters.

This fix specifically addresses:
- Chinese IME input (Pinyin, Wubi, etc.)
- Japanese IME input (Romaji, Kana, etc.)
- Korean IME input (Hangul, etc.)

## Tests

Added tests in `crates/term_input/src/tests.rs`:
- `test_ime_chinese_char`: Tests single Chinese character input
- `test_ime_esc_followed_by_chinese_simulation`: Simulates real IME input flow
- `test_ime_multiple_chinese_chars`: Tests multiple Chinese characters
- `test_alt_ascii_still_works`: Ensures Alt+ASCII still works
- `test_alt_non_ascii_still_works`: Ensures Alt+non-ASCII behavior is correct

## Related

This was split from PR #464 as requested by @osa1, since IME input handling is unrelated to the wide character display issue (#379).


osa1 (commented):
You moved some of the code that needed to stay in the other PR, to this PR..

There should be two PRs:

- One that deals with rendering wide characters. That's the other PR.
- One that deals with entering unicode input. That's this PR.

Can you separate these two? The first PR shouldn't have any input handling changes, this PR shouldn't have any rendering changes.

I'm inclined to just close your PRs as they're basically AI slop..

